PR #11, opened 2017-12-09
add constants and binary utils for binary trie

NIC619:
### How was it fixed?
Add binary trie part 1: Add the constants and the binary encoding/decoding utilities(and tests) for binary trie

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/trie/utils/binaries.py
+++ b/trie/utils/binaries.py
@@ -1,0 +1,66 @@
+from eth_utils import (
+    apply_to_return_value,
+)
+from cytoolz import (
+    partition_all,
+)
+
+
+from trie.constants import (
+    EXP,
+    TWO_BITS,
+    PREFIX_00,
+    PREFIX_100000,
+)
+
+
+@apply_to_return_value(bytes)
+def decode_from_bin(input_bin):
+    """
+    0100000101010111010000110100100101001001 -> ASCII
+    """
+    for chunk in partition_all(8, input_bin):
+        yield sum(
+            2**exp * bit
+            for exp, bit
+            in enumerate(reversed(chunk))
+        )
+
+
+@apply_to_return_value(bytes)
+def encode_to_bin(value):
+    """
+    ASCII -> 0100000101010111010000110100100101001001
+    """
+    for char in value:
+        for exp in EXP:
+            if char & exp:
+                yield True
+            else:
+                yield False
+
+
+def encode_from_bin_keypath(input_bin):
+    """
+    Encodes a sequence of 0s and 1s into tightly packed bytes
+    Used in encoding key path of a KV-NODE
+    """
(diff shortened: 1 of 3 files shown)


trie/utils/binaries.py
```diff
@@ -0,0 +1,66 @@
+from trie.constants import (
+    TWO_BITS,
+    PREFIX_00,
+    PREFIX_100000,
+)
+
+
+def safe_ord(value):
+    if isinstance(value, int):
```
pipermerriam: FYI, `isinstance(value, int)` will return truthy for `True` and `False`.  You can use `eth_utils.is_integer` if you need that kind of precision.

NIC619:
@pipermerriam Thanks for the reminder. Replaced it with `is_integer` as a precaution.

trie/utils/binaries.py
```diff
@@ -0,0 +1,66 @@
...
+        return value
+    else:
+        return ord(value)
+
+
+def decode_from_bin(input_bin):
```
pipermerriam: I just toyed with an alternate implementation of this.

```python
from eth_utils import (
    apply_to_return_value,
)
from cytoolz import (
    partition_all,
)


@apply_to_return_value(bytes)
def alt_decode_from_bin(input_bin):
    for chunk in partition_all(8, input_bin):
        yield sum(
            2**exp * bit
            for exp, bit
            in enumerate(reversed(chunk))
        )
```

Some *very* rudimentary benchmarking suggests it's about 30% faster than the current implementation.

trie/utils/binaries.py
```diff
@@ -0,0 +1,66 @@
...
+    output = bytearray(len(input_bin) // 8)
+    for i in range(0, len(input_bin), 8):
+        byte_value = 0
+        for bit_value in input_bin[i:i+8]:
+            byte_value = byte_value * 2 + bit_value
+        output[i//8] = byte_value
+    return bytes(output)
+
+
+def encode_to_bin(value):
```
pipermerriam: Ended up also taking a stab at an alternate implementation of this.

```python
EXP = tuple(reversed(tuple(2**i for i in range(8))))


@apply_to_return_value(bytes)
def alt_encode_to_bin(value):
    for char in value:
        for exp in EXP:
            if char & exp:
                yield True
            else:
                yield False
```

Benchmarking suggests it's 5-20% faster than the current implementation and I think it's a bit easier to read as well.

pipermerriam (changes requested):
I got curious and ended up trying out some ideas that I think improve both readability *and* performance.  Lemme know what you think.

pipermerriam (approved):

NIC619:
Though I didn't go too much into performance part, it definitely looks much cleaner. Will definitely try to adhere to this pattern.

NIC619:
Merged with new binary encoding/decoding.

PR #12, opened 2017-12-10
add node parsing/encoding/decoding utils and tests

NIC619:
### How was it fixed?
Add binary trie node parsing/encoding/decoding utilities and tests

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/tests/test_nodes_utils.py
+++ b/tests/test_nodes_utils.py
@@ -1,8 +1,16 @@
 import pytest
 
+from trie.exceptions import (
+    InvalidNode,
+    ValidationError,
+)
 from trie.utils.nodes import (
     get_common_prefix_length,
     consume_common_prefix,
+    parse_node,
+    encode_kv_node,
+    encode_branch_node,
+    encode_leaf_node,
 )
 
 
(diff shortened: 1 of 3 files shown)


tests/test_nodes_utils.py
```diff
@@ -40,3 +48,71 @@ def test_consume_common_prefix(left, right, expected):
...
+        (b"\x00\xc5\xd2F\x01\x86\xf7#<\x92~}\xb2\xdc\xc7\x03\xc0\xe5\x00\xb6S\xca\x82';{\xfa\xd8\x04]\x85\xa4p"),  # noqa: E501
+        (b"\x01\xc5\xd2F\x01\x86\xf7#<\x92~}\xb2\xdc\xc7\x03\xc0\xe5\x00\xb6S\xca\x82';{\xfa\xd8\x04]\x85\xa4p"),  # noqa: E501
+        (b"\x01\x02\xc5\xd2F\x01\x86\xf7#<\x92~}\xb2\xdc\xc7\x03\xc0\xe5\x00\xb6S\xca\x82';{\xfa\xd8\x04]\x85\xa4p\xc5\xd2F\x01\x86\xf7#<\x92~}\xb2\xdc\xc7\x03\xc0\xe5\x00\xb6S\xca\x82';{\xfa\xd8\x04]\x85\xa4p"),  # noqa: E501
+        (b'\x02'),
+    ),
+)
+def test_binary_trie_node_parsing_with_invalid_input(node):
+    try:
+        parse_node(node)
+    except InvalidNode:
```
pipermerriam: This test would currently pass if any of the test cases above did not raise `InvalidNode`

The way to test this type of thing using pytest is using `pytest.raises`

```python
with pytest.raises(InvalidNode):
    parse_node(node)
```

Tests written this way will fail if the provided exception is not raised.

tests/test_nodes_utils.py
```diff
@@ -40,3 +48,71 @@ def test_consume_common_prefix(left, right, expected):
...
+        (b"\xc5\xd2F\x01\x86\xf7#<\x92~}\xb2\xdc\xc7\x03\xc0\xe5\x00\xb6S\xca\x82';{\xfa\xd8\x04]\x85\xa4p", b'\x01'),  # noqa: E501
+        (b"\xc5\xd2F\x01\x86\xf7#<\x92~}\xb2\xdc\xc7\x03\xc0\xe5\x00\xb6S\xca\x82';{\xfa\xd8\x04]\x85\xa4p", 12345),  # noqa: E501
+        ([3] * 32, [4] * 32),
+        (b'\x01' * 33, b'\x01' * 32),
+    ),
+)
+def test_encode_binary_trie_branch_node_with_invalid_input(left_child_node_hash, right_child_node_hash):
+    try:
+        encode_branch_node(left_child_node_hash, right_child_node_hash)
+    except ValidationError:
```
pipermerriam: Same note about how to test exception raising.

tests/test_nodes_utils.py
```diff
@@ -40,3 +48,71 @@ def test_consume_common_prefix(left, right, expected):
...
+    (
+        (b''),
+        (12345),
+        (range(5)),
+    ),
+)
+def test_encode_binary_trie_leaf_node_with_invalid_input(value):
+    try:
+        encode_leaf_node(value)
+    except ValidationError:
```
pipermerriam: Same note about exception raising tests.

trie/utils/nodes.py
```diff
@@ -89,3 +101,62 @@ def consume_common_prefix(left_key, right_key):
...
+    """
+    Input: a serialized node
+    """
+    if node is None or node == b'':
+        raise InvalidNode("Blank node is not a valid node type in Binary Trie")
+    elif node[0] == BRANCH_TYPE:
+        if len(node) != 65:
+            raise InvalidNode("Invalid branch node, both child node should be 32 bytes long each")
+        # Output: left child, right child, node type
+        return node[1:33], node[33:], BRANCH_TYPE
```
pipermerriam: I'm curious about the choice to return the type as the 3rd parameter.  Maybe my intuition is off but I would have expected it to be the first parameter.  Not sure why, but it feels more correct.  

Thoughts? (I'm not saying this is wrong, just curious if there is a reason for the current ordering)

trie/utils/nodes.py
```diff
@@ -89,3 +101,62 @@ def consume_common_prefix(left_key, right_key):
...
+def encode_kv_node(keypath, child_node_hash):
+    """
+    Serializes a key/value node
+    """
+    if keypath is None:
+        raise ValidationError("Key path can not be empty")
+    validate_is_bytes(keypath)
+    validate_is_bytes(child_node_hash)
+    validate_length(child_node_hash, 32)
+    return bytes([KV_TYPE]) + encode_from_bin_keypath(keypath) + child_node_hash
```
pipermerriam: We can probably gain some *minor* performance improvements if we stored `KV_NODE_PREFIX = bytes([KV_TYPE])` as a constant.

Same for the other prefixes.

pipermerriam (changes requested):
I think it would be prudent to also include tests for the non-failure cases for each of these utilities.  They serve as both a spec for future contributors to the codebase as well as simply ensuring that these tools work for the basic use cases we expect.

NIC619:
1. added test cases with valid input
or they can be combined into one test case like
```
@pytest.mark.parametrize(
    'valid_node,invalid_node',
    (
        (b"\x02value", b"\x02"),
    ),
)
def test_binary_trie_node_parsing(valid_node, invalid_node)
```
for cleanliness

2. added constant node type prefix and change arrangement of returned data of `parse_node`

tests/test_nodes_utils.py
```diff
@@ -40,3 +48,105 @@ def test_consume_common_prefix(left, right, expected):
...
+@pytest.mark.parametrize(
+    'valid_node',
+    (
+        (b"\x00\x03\x04\x05\xc5\xd2F\x01\x86\xf7#<\x92~}\xb2\xdc\xc7\x03\xc0\xe5\x00\xb6S\xca\x82';{\xfa\xd8\x04]\x85\xa4p"),  # noqa: E501
+        (b"\x01\xc5\xd2F\x01\x86\xf7#<\x92~}\xb2\xdc\xc7\x03\xc0\xe5\x00\xb6S\xca\x82';{\xfa\xd8\x04]\x85\xa4p\xc5\xd2F\x01\x86\xf7#<\x92~}\xb2\xdc\xc7\x03\xc0\xe5\x00\xb6S\xca\x82';{\xfa\xd8\x04]\x85\xa4p"),  # noqa: E501
+        (b"\x02value"),  # noqa: E501
+    ),
+)
+def test_binary_trie_node_parsing_with_valid_input(valid_node):
+    parse_node(valid_node)
```
pipermerriam: What do you think about adding assertions that the return of this function is returning the expected values?  Seems like a small addition that would improve this test quite a bit.

tests/test_nodes_utils.py
```diff
@@ -40,3 +48,105 @@ def test_consume_common_prefix(left, right, expected):
...
+
+
+@pytest.mark.parametrize(
+    'keypath,valid_value',
+    (
+        (b'\x00', b"\xc5\xd2F\x01\x86\xf7#<\x92~}\xb2\xdc\xc7\x03\xc0\xe5\x00\xb6S\xca\x82';{\xfa\xd8\x04]\x85\xa4p"),  # noqa: E501
+    ),
+)
+def test_encode_binary_trie_kv_node_with_valid_input(keypath, valid_value):
+    encode_kv_node(keypath, valid_value)
```
pipermerriam: Same suggestion here about adding some assertions about the value returned from this function.

tests/test_nodes_utils.py
```diff
@@ -40,3 +48,105 @@ def test_consume_common_prefix(left, right, expected):
...
+
+
+@pytest.mark.parametrize(
+    'left_child_node_hash,right_child_node_hash',
+    (
+        (b'\xc8\x9e\xfd\xaaT\xc0\xf2\x0cz\xdfa(\x82\xdf\tP\xf5\xa9Qc~\x03\x07\xcd\xcbLg/)\x8b\x8b\xc6', b"\xc5\xd2F\x01\x86\xf7#<\x92~}\xb2\xdc\xc7\x03\xc0\xe5\x00\xb6S\xca\x82';{\xfa\xd8\x04]\x85\xa4p"),  # noqa: E501
+    ),
+)
+def test_encode_binary_trie_branch_node_with_valid_input(left_child_node_hash, right_child_node_hash):
+    encode_branch_node(left_child_node_hash, right_child_node_hash)
```
pipermerriam: Same here about adding assertions for the return value of this function

tests/test_nodes_utils.py
```diff
@@ -40,3 +48,105 @@ def test_consume_common_prefix(left, right, expected):
...
+
+
+@pytest.mark.parametrize(
+    'valid_value',
+    (
+        (b'\x01\x02x\03'),
+    ),
+)
+def test_encode_binary_trie_leaf_node_with_invalid_input(valid_value):
+    encode_leaf_node(valid_value)
```
pipermerriam: Same here about adding assertions for the return value of this function

pipermerriam (approved):
RE the comments suggesting improvements to the tests.

If you are in a hurry and need to move this forward, feel free to add some `TODO` comments to the code and open up an corresponding issue on github indicating that this test suite can be further fleshed out.

NIC619:
@pipermerriam test suites updated. Requesting a review. Thanks

pipermerriam (approved):

PR #14, opened 2017-12-13
Add BinaryTrie class and one test suite

NIC619:
1. Make the BinaryTrie looked as close to original Trie as possible so user won't have much trouble switching between two different tries.
2. Separate branch related functions into branch utilities which will be included in next PR.
3. Need to add more test suites.

#### Cute Animal Picture

Cute animal picture


NIC619:
@pipermerriam should I split them into smaller PRs?

tests/test_bin_trie.py
```diff
@@ -0,0 +1,40 @@
...
+)
+
+
+kvpairs = [(keccak(str(i).encode('utf-8')), str(i).encode('utf-8') * 5) for i in range(2000)]
+
+
+def test_bin_trie_different_order_insert():
+    result = BLANK_HASH
+    # Repeat 3 times
+    for _ in range(3):
```
pipermerriam: Can we change the approach to this test.  What I'm worried about is this finding an edge case that fails but not actually knowing what it was because it involves some unknown random ordering.

I suggest we do the following.

- convert to use `hypothesis` for randomness generation.

```
from hypothesis import (
    given,
    strategies as st,
)

@given(kv_pairs=st.lists(st.tuples(st.binary(), st.binary())))
def test_bin_trie_different_order_insert(kv_pairs):
    ...
```

Then, to check that ordering is unimportant we still need randomness.  `hypothesis` can help with this too via `hypothesis.strategies.random_module()` which returns the equivalent of `import random` but with the seed set by hypothesis so that the test can be repeated deterministically. 

```
from hypothesis import (
    given,
    strategies as st,
)

@given(kv_pairs=st.lists(st.tuples(st.binary(), st.binary())), random=st.random_module())
def test_bin_trie_different_order_insert(kv_pairs, random):
    ....
    random.shuffle(kv_pairs)
    ...
```

trie/trie.py
```diff
@@ -378,3 +397,250 @@ def __delitem__(self, key):
 
     def __contains__(self, key):
         return self.exists(key)
+
+
+class BinaryTrie():
```
pipermerriam: can you add `object` to this as the base class.

trie/trie.py
```diff
@@ -378,3 +397,250 @@ def __delitem__(self, key):
...
+        Sets the value at the given keypath from the given node
+
+        Keyp will be encoded into binary array format first
+        """
+        validate_is_bytes(key)
+        validate_is_bytes(value)
+
+        self.root_hash = self._set(self.root_hash, encode_to_bin(key), value)
+
+    def _set(self, node_hash, keypath, value):
```
pipermerriam, line 458: This function body is about 150 lines long.

Can we split it up a bit.  Maybe split the body up into individual functions for each node type to reduce the size a bit?

```python
if nodetype == LEAF_TYPE:
    self._set_leaf_node(...)
elif nodetype == KV_TYPE:
    self._set_kv_node(...)
elif nodetype == BRANCH_TYPE:
    self._set_branch_node(...)
```

trie/validation.py
```diff
@@ -38,3 +42,10 @@ def validate_is_node(node):
                 validate_length(sub_node, 32)
     else:
         raise ValidationError("Invalid Node: {0}".format(node))
+
+
+def validate_is_bin_node(node):
+    if node == BLANK_HASH or node[0] == BRANCH_TYPE or node[0] == KV_TYPE or node[0] == LEAF_TYPE:
```
pipermerriam: if `node[0] in {BRANCH_TYPE, KV_TYPE, LEAF_TYPE}` is a bit more concise way to express this logic.

pipermerriam (changes requested):

tests/test_bin_trie.py
NIC619: some limits with random kv pairs are:
1. in binary trie, one key can not be the prefix of the other key, e.g, b'\x00\x01' and b'\x00\x01\x02'
2. can not have updates with different values but the same key, otherwise it might result in different root hash after shuffling

so I make some twists:
```
@to_list
def to_kv_pairs(kv_zip):
    for k, v in kv_zip:
        yield (k, v)

@given(k=st.lists(st.binary(min_size=32, max_size=32), min_size=10, max_size=10, unique=True),
    v=st.lists(st.binary(min_size=1), min_size=10, max_size=10),
    set_random=st.random_module())
def test_bin_trie_different_order_insert(k, v, set_random):
    kv_pairs = to_kv_pairs(zip(k, v))
```
Which basically make all keys to be 32 bytes long and values to be 1 bytes minimum.
Numbers including list size are all adjustable.
What do you think?

pipermerriam: Looks like a solid approach. 👍

NIC619:
@pipermerriam I updated the trie and the test suite and it's ready for review. For the test suite, with default settings in hypothesis it could take tens of seconds to run the test so I set the `max_examples` to 10 for now.

tests/test_bin_trie.py
```diff
@@ -0,0 +1,51 @@
...
+
+from trie.trie import (
+    BinaryTrie,
+)
+from trie.constants import (
+    BLANK_HASH,
+)
+
+@to_list
+def to_kv_pairs(kv_zip):
```
pipermerriam: I think this function can be replaced with a call to `list` or `tuple` to simply evaluate the underlying `zip` call (unless I'm missing something.)

tests/test_bin_trie.py
```diff
@@ -0,0 +1,51 @@
...
+        for i, (k, v) in enumerate(kv_pairs):
+            trie.set(k, v)
+            assert trie.get(k) == v
+        assert result is BLANK_HASH or trie.root_hash == result
+        result = trie.root_hash
+        # insert already exist key/value
+        trie.set(kv_pairs[0][0], kv_pairs[0][1])
+        assert trie.root_hash == result
+        # Delete all key/value
+        random.shuffle(kv_pairs)
```
pipermerriam, line 35: this should be using `set_random.shuffle(...)`

tests/test_bin_trie.py
```diff
@@ -0,0 +1,51 @@
...
+    v=st.lists(st.binary(min_size=1), min_size=100, max_size=100),
+    set_random=st.random_module())
+@settings(max_examples=10)
+def test_bin_trie_different_order_insert(k, v, set_random):
+    kv_pairs = to_kv_pairs(zip(k, v))
+    result = BLANK_HASH
+    # Repeat 3 times
+    for _ in range(3):
+        trie = BinaryTrie(db={})
+        random.shuffle(kv_pairs)
```
pipermerriam, line 25: I think this should be using `set_random.shuffle()`.  Even if the `set_random` fixture appropriately patches things in the `random` module, it still feels appropriate to use the fixture version of the random module so that it's clear that we're using deterministic randomness.

tests/test_bin_trie.py
```diff
@@ -0,0 +1,51 @@
...
+        # insert already exist key/value
+        trie.set(kv_pairs[0][0], kv_pairs[0][1])
+        assert trie.root_hash == result
+        # Delete all key/value
+        random.shuffle(kv_pairs)
+        for k, v in kv_pairs:
+            trie.delete(k)
+        assert trie.root_hash == BLANK_HASH
+
+# TODO: add more binary trie tests
```
pipermerriam: Are there specific tests you think should be added?  If so, can you remove this comment and open a github issue for them?

trie/trie.py
```diff
@@ -378,3 +397,256 @@ def __delitem__(self, key):
...
+            # into one node
+            if subnodetype == KV_TYPE:
+                return self._hash_and_save(
+                    encode_kv_node(left_child + sub_left_child, sub_right_child)
+                )
+            else:
+                return self._hash_and_save(encode_kv_node(left_child, subnode_hash))
+        # Keypath prefixes don't match. Here we will be converting a key-value node
+        # of the form (k, CHILD) into a structure of one of the following forms:
+        # i.   (k[:-1], (NEWCHILD, CHILD))
```
pipermerriam: Can we re-label these cases to either:

- `1, 2, 3, 4`
- `A, B, C, D`

Those are both easier for me to visually differentiate.

trie/trie.py
```diff
@@ -378,3 +397,256 @@ def __delitem__(self, key):
...
+                    encode_kv_node(left_child[:common_prefix_len], newsub)
+                )
+            # Case 2: keypath prefixes diverge in the first bit, so we replace the
+            # kv node with a branch node
+            # (v) (vi) (vii) (viii)
+            else:
+                return newsub
+
+    def _set_branch_node(self, keypath, node_type, left_child, right_child, value):
+        new_left_child, new_right_child = left_child, right_child
```
pipermerriam: I don't love overriding variables when possible.  What do you think about changing the `if/else` below to something like the following

```python
        if keypath[:1] == BYTE_0:
            new_left_child = self._set(left_child, keypath[1:], value)
            new_right_child = right_child
        else:
            new_right_child = self._set(right_child, keypath[1:], value)
            new_left_child = left_child
```

trie/trie.py
```diff
@@ -378,3 +397,256 @@ def __delitem__(self, key):
...
+                        new_left_child
+                        if new_left_child != BLANK_HASH
+                        else new_right_child
+                    )
+                )
+        else:
+            return self._hash_and_save(encode_branch_node(new_left_child, new_right_child))
+
+    def exists(self, key):
+        validate_is_bytes(key)
```
pipermerriam: You should be able to remove this validation check since it occurs within `self.get(...)`

trie/trie.py
```diff
@@ -378,3 +397,256 @@ def __delitem__(self, key):
...
+    #
+    # Convenience
+    #
+    @property
+    def root_node(self):
+        return self.db[self.root_hash]
+
+    @root_node.setter
+    def root_node(self, node):
+        validate_is_bin_node(node)
```
pipermerriam, line 620: You should be able to remove this check since it occurs within `_hash_and_save(...)`

trie/validation.py
```diff
@@ -38,3 +42,10 @@ def validate_is_node(node):
                 validate_length(sub_node, 32)
     else:
         raise ValidationError("Invalid Node: {0}".format(node))
+
+
+def validate_is_bin_node(node):
+    if node == BLANK_HASH or node[0] in (BRANCH_TYPE, KV_TYPE, LEAF_TYPE):
```
pipermerriam: *small nitpick*

Lets move `(BRANCH_TYPE, KV_TYPE, LEAF_TYPE)` to a constant, something like `BINARY_TREE_NODE_TYPES` and change it to be a set.

Then lets expand this function for slightly easier readability.

```python
if node = BLANK_HASH:
    ....
elif node in BINARY_TREE_NODE_TYPES:
    ...
else:
    ...
```

pipermerriam (approved):
I left some suggested changes for final cleanup.  This looks good to go otherwise.  :shipit: when you're happy with it unless you make sufficient changes you'd like me to look over it again.

Also, It might be good to update the README but only if you're in a place that you're ready to add basic documentation (feel free to do this in another pull request)

tests/test_bin_trie.py
NIC619: After look into hypothesis documents, it seems that `strategy.random_module` seed the `random` module at the start of the test and what it returns is an object that basically represents the seed (and hence no `shuffle` function) while `strategy.randoms` returns a random number generator. I will change it to `strategy.randoms` since the code does not depend on the global`random` module.

tests/test_bin_trie.py
NIC619: Currently have one or two edge cases in mind but generally hope there could be more tests(since tests suites for `HexaryTrie` is not applicable to `BinaryTrie`). To make sure it's as robust as `HexaryTrie`.

tests/test_bin_trie.py
pipermerriam: So it is.  Sorry to have mislead you.  Carry on!

PR #15, opened 2017-12-20
New verify_proof() function

gsalgado:


trie/proof.py
```diff
@@ -0,0 +1,23 @@
...
+
+    Returns the value for the key when it is present, an empty bytes when the key is not present,
+    or raises BadTrieProof if the proof is not valid.
+    """
+    trie = Trie({})
+    for node in proof:
+        trie._persist_node(node)
+    trie.root_hash = root_hash
+    try:
+        return trie.get(key)
```
Matthalp-zz, line 21: Nit: you could use`trie.exists(key)`.

gsalgado: I want to return the key's value here, not just check if it exists

Matthalp-zz: I misread the pydoc -- my mistake.

trie/proof.py
```diff
@@ -0,0 +1,23 @@
...
+    except KeyError as e:
```
Matthalp-zz, line 22: Just for my understanding: a `BadTrieProof` will only be raised when the nodes are malformed?

trie/proof.py
```diff
@@ -0,0 +1,23 @@
...
+
+from trie.exceptions import BadTrieProof
+
+
+# TODO: This function needs to be audited to ensure the approach used here to verify a trie proof
+# is correct.
+def verify_proof(root_hash, key, proof):
+    """Verify that the given proof contains a value for key on a trie with the given root hash.
+
+    Returns the value for the key when it is present, an empty bytes when the key is not present,
```
Matthalp-zz, line 13: Not a big deal but typically when I see a `verify` function, I expect it to return a boolean based on whether or not the verification was successful. I'm not sure how important it is for you to recover the actual value so I'll leave that up to you. It might be overkill but if you need both you could have a`Proof` object that provides each of these separately, which looks like the `Optional` type in other languages.

trie/proof.py
gsalgado: When nodes along the key's path are missing

Matthalp-zz: My understanding is that if a node along the key's path is missing then a `BLANK_NODE` will eventually be returned.

trie/proof.py
Matthalp-zz, line 21: It looks like `trie.get` returns the node at that path and not the value. I think you will have to check the resulting node to determine what the value actually is.

gsalgado: Nope, it returns the value

Matthalp-zz: You are right: https://github.com/ethereum/py-trie/blob/master/trie/trie.py#L357

trie/proof.py
```diff
@@ -0,0 +1,23 @@
+from __future__ import absolute_import
+
+from trie import Trie
+
+from trie.exceptions import BadTrieProof
+
+
+# TODO: This function needs to be audited to ensure the approach used here to verify a trie proof
+# is correct.
+def verify_proof(root_hash, key, proof):
```
pipermerriam, line 10: What do you think about exposing this as a `classmethod` on the `Trie` class?

pipermerriam (approved):
TLDR: this is good to go with the removal of the bumpversion commit which I'll handle.

I'm still a little nervous about this, mostly because I still don't have a full enough understanding of *why* this works to be able to explore edge cases where it could fail.  

The approach seems both elegant and potentially dangerous since the verification is passive.  I'd like to migrate this to a more direct approach at some point if that is feasible so that our proof verification is not dependent on the workings of internal APIs.

pipermerriam:
replaced by #17

PR #19, opened 2017-12-21
 add trie.delete_subtrie and if_branch_exist utility function

NIC619:
### How was it fixed?
Add `trie.delete_subtrie`, which is needed in py-evm Account Redesign, and some tests.
Renamed `delete_branch` to `delete_subtrie` for that `delete_subtrie` is closer to it's actual effect.

Will update `if_branch_exist` and tests soon. `if_branch_exist` will be placed in `trie/utils/branches.py` which is where utility functions for trie branch are stored.
There will be more branch utils and their tests added in next PR. 

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/tests/test_bin_trie.py
+++ b/tests/test_bin_trie.py
@@ -1,3 +1,5 @@
+import pytest
+
 from hypothesis import (
     given,
     strategies as st,
@@ -9,6 +11,9 @@ from trie.trie import (
 )
 from trie.constants import (
     BLANK_HASH,
+)
+from trie.exceptions import (
+    LeafNodeOverrideError,
 )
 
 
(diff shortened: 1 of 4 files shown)


NIC619:
@pipermerriam add `if_branch_exist` and tests, please have a look.

I plan to make branch utility functions e.g, get/verify merkle branch and get subtrie/witness, as standalone pure functions so we don't need to go through the procedure of setting up a trie every time we need to fetch a branch or witness. Just feed the function with trie database and trie root will do the job.

tests/test_bin_trie.py
```diff
@@ -36,3 +41,44 @@ def test_bin_trie_different_order_insert(k, v, random):
         for k, v in kv_pairs:
             trie.delete(k)
         assert trie.root_hash == BLANK_HASH
+
+
+def test_bin_trie_delete_subtrie():
```
pipermerriam: Can you refactor this test to use `pytest.mark.parametrize` so that we don't end up with any accidental leakage between the different test cases.

tests/test_branches_utils.py
```diff
@@ -0,0 +1,33 @@
...
+
+from trie.trie import (
+    BinaryTrie,
+)
+from trie.utils.branches import (
+    if_branch_exist,
+)
+
+
+def set_up_test_trie():
```
pipermerriam: *small refactor suggestion*

If you made this into a fixture:

```python
@pytest.fixture
def trie(...):
    ...
```

Then you could just include it in the arguments of `test_branch_exist` and use it directly as it will be reset for each test.

trie/trie.py
```diff
@@ -455,7 +455,7 @@ def set(self, key, value):
 
         self.root_hash = self._set(self.root_hash, encode_to_bin(key), value)
 
-    def _set(self, node_hash, keypath, value):
+    def _set(self, node_hash, keypath, value, if_delete_subtrie=False):
```
pipermerriam, line 458: I'm confused by the name `if_delete_subtrie`.  It's not clear what it's signaling.  Can you explain or add a comment explaining in human-readable-text what it does.

trie/utils/branches.py
```diff
@@ -0,0 +1,46 @@
...
+)
+from trie.utils.binaries import (
+    encode_to_bin,
+)
+from trie.utils.nodes import (
+    parse_node,
+)
+
+
+def if_branch_exist(db, node_hash, key_prefix):
```
pipermerriam, line 16: Is this intended to be a public API?  If so, it should be moved outside of `utils` as `utils` is intended to be purely private APIs

pipermerriam (approved):
Looks good.

Can you get an update to the README in to fill in some documentation for this new trie?

pipermerriam:
:shipit: when ready.  I'll cut you a release.

trie/trie.py
NIC619: It's used by `delete_subtrie` and call from `delete_subtrie` will set this flag to `True` which indicates if this `_set` action should delete subtrie of a given prefix instead of setting a new value for a given key.

`delete_subtrie` can be written as a standalone function but since it shares most of the logic with `_set`, I reuse the code from `_set` and add a flag to indicate different actions.

I will add a comment for it.

trie/utils/branches.py
NIC619: yes it's intended to be a public API. I will move it to `trie/branches.py`.

NIC619:
My schedule now is roughly
1. branch functions and tests
-> 2. Trie rename, Trie subclass and deprecation warning
-> 3. documentation

If documentation is needed now I can do it after 1.

pipermerriam:
As long as the docs get written I don't mind what order.  Would you like a release with these latest changes or would you rather wait until the other changes have been applied?

NIC619:
I can come up with documents for these branch functions in a couple days. Then add some basic introduction to BinaryTrie along with HexaryTrie renaming and deprecation warning in next release?

PR #20, opened 2017-12-26
add branch/witness helper functions

NIC619:
1. move `branches.py` from `trie/utils/` to `trie/`
2. add `get_branch` and `verify_branch`
3. add `get_trie_nodes` which returns all trie nodes given a root node hash
4. add `get_witness` which returns witness given a key path
5. test suites

PS: note sure if the name `branches` is intuitive enough
#### Cute Animal Picture

Cute animal picture


trie/branches.py
```diff
@@ -0,0 +1,167 @@
...
+    elif nodetype == BRANCH_TYPE:
+        if not keypath:
+            raise InvalidKeyError("Key too short")
+        if keypath[:1] == BYTE_0:
+            return [node] + _get_branch(db, left_child, keypath[1:])
+        else:
+            return [node] + _get_branch(db, right_child, keypath[1:])
+
+
+def verify_branch(branch, root_hash, key, value):
```
pipermerriam: *naming nitpick*

For things that raise exceptions, I've used the convention `verify_thing` or `validate_thing`

For things that return `True/False` I've tended towards the conventions:

- `check_if_thing`
- `check_thing`
- `is_thing`
- `is_thing_valid`

I would suggest the name `is_branch_valid` for this or something else along the same naming convention.

trie/branches.py
```diff
@@ -0,0 +1,167 @@
...
+    validate_is_bytes(key)
```
pipermerriam: This validation can be removed since it occurs within `BinaryTrie.get(...)`.

However, I think that it might be worth validating that `branch`, `value` are well formed since they don't get validated anywhere.  (`root_hash` is validated with `BinaryTrie.__init__(...)`

trie/branches.py
```diff
@@ -0,0 +1,167 @@
...
+        node = db[node_hash]
+    else:
+        return []
+    nodetype, left_child, right_child = parse_node(node)
+    if nodetype == KV_TYPE:
+        return [node] + get_trie_nodes(db, right_child)
+    elif nodetype == BRANCH_TYPE:
+        return [node] + get_trie_nodes(db, left_child) + get_trie_nodes(db, right_child)
+    elif nodetype == LEAF_TYPE:
+        return [node]
```
pipermerriam: Can you add a final `else` with a `raise Exception("Invariant: unreachable code path")` just for good measure.

trie/branches.py
```diff
@@ -0,0 +1,167 @@
...
+            return [node] + get_trie_nodes(db, right_child)
+        if keypath[:len(left_child)] == left_child:
+            return [node] + _get_witness(db, right_child, keypath[len(left_child):])
+        else:
+            raise InvalidKeyError("Key does not exist")
+    elif nodetype == BRANCH_TYPE:
+        if keypath[:1] == BYTE_0:
+            return [node] + _get_witness(db, left_child, keypath[1:])
+        else:
+            return [node] + _get_witness(db, right_child, keypath[1:])
```
pipermerriam: Can you add a final `else` with a `raise Exception("Invariant: unreachable code path")` just for good measure.

trie/branches.py
```diff
@@ -0,0 +1,167 @@
...
+        return []
+    nodetype, left_child, right_child = parse_node(node)
+    if nodetype == LEAF_TYPE:
+        if keypath:
+            raise InvalidKeyError("Key too long")
+    elif nodetype == KV_TYPE:
+        if len(keypath) < len(left_child) and left_child[:len(keypath)] == keypath:
+            return [node] + get_trie_nodes(db, right_child)
+        if keypath[:len(left_child)] == left_child:
+            return [node] + _get_witness(db, right_child, keypath[len(left_child):])
```
pipermerriam: I think we can achieve some minor performance gains here if we switch this function to use `yield` statements.  

```python
return [node] + get_trie_nodes(...)

# becomes...
yield node
yield from _get_witness(...)
```

Then, in the main `get_witness` function, you can change the return statement to `return tuple(_get_witness(...))`.  This removes the requirement for python to create multiple lists and then repeatedly concatenate them.

I think this same pattern can also be applied to `get_trie_nodes` too.

pipermerriam (changes requested):

trie/branches.py
NIC619: There are a few situations I'm not sure if we should check for.
1. node data not in db => a. a dictionary `KeyError` or b. check if key exist first and raise `NodeNotExistError`?
1.b looks like a more clear error message than 1.a to me

2. get branch from an empty trie => a. return empty branch or b. raise trying to fetch from empty trie error?
Since fetching non-exist key from empty trie returns value `None`, 2.a is probably a more consistent approach with this pattern. Or we can implement them both as 2.b. The question I think is, should the user be awarded of that they are fetching values from an empty trie. (In HexaryTrie it will return `BLANK_NODE` but in BinaryTrie there's no `BLANK_NODE`. And no node hash should be `BLANK_HASH` except root hash.) If we implement 2.a then an empty branch would be a valid branch which seems weird to me but has no significant side effect in my knowledge.

NIC619: On a second thought it's probably not a good idea to raise error while fetching value from empty trie. I think I'm just going to treat empty branch as a valid one.

pipermerriam: I'm not familiar enough with the binary trie spec to know which of these is right.  

It sounds like treating empty branches as valid violates the [Principle of Least Surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) indicating that raising an exception in this case would be the expected behavior.

trie/branches.py
```diff
@@ -0,0 +1,190 @@
...
+        validate_is_bytes(key)
+    for node in branch:
+        validate_is_bin_node(node)
+
+    db = {keccak(node): node for node in branch}
+    assert BinaryTrie(db=db, root_hash=root_hash).get(key) == value
+    return True
+
+
+def get_trie_nodes(db, node_hash):
```
pipermerriam, line 129: It's worth noting that this function will return a generator which is probably not what you actually want.  I would suggest writing an outer function `get_trie_nodes` and renaming this function to `_get_trie_nodes` like the following.


```python
def get_trie_nodes(...):
    return tuple(_get_trie_nodes(...))
```

This will allow all of the various iterators to be lazily evaluated at the very end and cast to a single tuple.

trie/branches.py
```diff
@@ -0,0 +1,190 @@
...
+        raise Exception("Invariant: unreachable code path")
+
+
+def get_branch(db, root_hash, key):
+    """
+    Get a long-format Merkle branch
+    """
+    validate_is_bytes(key)
+
+    return list(_get_branch(db, root_hash, encode_to_bin(key)))
```
pipermerriam: *nitpick*

Any reason for using a `list` over a `tuple`?  The inherent immutability of `tuples` is nice imho.

trie/branches.py
```diff
@@ -0,0 +1,190 @@
...
+    """
+    Given a key prefix, return whether this prefix is
+    the prefix of an existing key in the trie.
+    """
+    validate_is_bytes(key_prefix)
+
+    return _if_branch_exist(db, root_hash, encode_to_bin(key_prefix))
+
+
+def _if_branch_exist(db, node_hash, key_prefix):
```
pipermerriam: *naming nitpick*

`check_if_branch_exists` might convey the functionality of this function a bit better.

pipermerriam (approved):
Some basic cleanup if your

pipermerriam:
@NIC619 let me know if you need/want a release with these changes in it.

trie/branches.py
```diff
@@ -0,0 +1,196 @@
...
+            yield node
+    elif nodetype == BRANCH_TYPE:
+        if keypath[:1] == BYTE_0:
+            yield node
+            yield from _get_witness(db, left_child, keypath[1:])
+        else:
+            yield node
+            yield from _get_witness(db, right_child, keypath[1:])
+    else:
+        raise Exception("Invariant: unreachable code path")
```
vbuterin, line 196: `get_witness` was renamed from `get_prefix_witness`, correct? I actually like the older naming better, because it reflects the fact that this method is supposed to return a witness that covers all key/value pairs that have a given prefix.

NIC619: My feeling was that `get_prefix_witness` might make people think that it's the *prefix of witness*. So I take *prefix* out of the name and explain it in docstring:
```
Get all witness given a keypath prefix.  Include
1. witness along the keypath and
2. witness in the subtrie of the last node in keypath
```

what about `get_witness_by_key_prefix`?

PR #25, opened 2018-01-06
some Binary Trie updates

NIC619:
### How was it fixed?
- [x] rename "get_witness" to "get_witness_for_key_prefix"
- [x] fix: have `get()` return None for key which is too long even if it's prefix is a valid key, e.g.,
if `trie.get(b'\x00\x11\x22')` returns `b'123'`, `trie.get(b'\x00\x11\x22\x33')` should return `None`
- [x] add invalid key `get`/`delete` test

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ False
 >>>     check_if_branch_exist,
 >>>     get_branch,
 >>>     if_branch_valid,
->>>     get_witness,
+>>>     get_witness_for_key_prefix,
 >>> )
 >>> t = BinaryTrie(db={})
 >>> t.root_hash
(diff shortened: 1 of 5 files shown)


pipermerriam (approved):
Can you explain why accessing a key which is too long but has a valid prefix should return `None`.

My naive intuition says that is should throw a `KeyError` and I'm curious why returning `None` is the appropriate behavior?

NIC619:
Since in BinaryTrie accessing key which is too short or which does not exist will return `None` and in HexaryTrie it returns `b''` for all three cases. So for consistency I treat this case the same as the other two.

PR #26, opened 2018-01-08
Implement HexaryTrieSync

gsalgado:
This is the Trie-related part of the code that will perform the state
trie download that we'll do when fast-syncing.

I was able to use this to sync the ropsten state trie, and that one test
gives us 88% code coverage, so I believe this is pretty solid

### DIFF
--- a/trie/hexary.py
+++ b/trie/hexary.py
@@ -25,13 +25,14 @@ from trie.utils.nibbles import (
     encode_nibbles,
 )
 from trie.utils.nodes import (
+    decode_node,
     get_node_type,
     extract_key,
     compute_leaf_key,
     compute_extension_key,
+    is_blank_node,
     is_extension_node,
     is_leaf_node,
-    is_blank_node,
     consume_common_prefix,
     key_starts_with,
 )
@@ -171,7 +172,7 @@ class HexaryTrie(object):
             encoded_node = node_hash
         else:
             encoded_node = self.db[node_hash]
-        node = self._decode_node(encoded_node)
+        node = decode_node(encoded_node)
 
         return node
 
@@ -186,14 +187,6 @@ class HexaryTrie(object):
         encoded_node_hash = keccak(encoded_node)
         self.db[encoded_node_hash] = encoded_node
         return encoded_node_hash
-
-    def _decode_node(self, encoded_node_or_hash):
-        if encoded_node_or_hash == BLANK_NODE:
-            return BLANK_NODE
-        elif isinstance(encoded_node_or_hash, list):
-            return encoded_node_or_hash
-        else:
-            return rlp.decode(encoded_node_or_hash)
 
     #
     # Node Operation Helpers
--- a/trie/sync.py
+++ b/trie/sync.py
@@ -1,0 +1,182 @@
+import bisect
+
+from eth_utils import encode_hex
+
+from trie.constants import (
+    NODE_TYPE_BRANCH,
+    NODE_TYPE_EXTENSION,
+    NODE_TYPE_LEAF,
+)
+from trie.exceptions import (
+    SyncRequestAlreadyProcessed,
+    UnexpectedNodeType,
+)
+from trie.utils.nodes import (
+    decode_node,
+    get_node_type,
+    is_blank_node,
+)
+from trie.utils.sha3 import keccak
+
+
+class SyncRequest:
+
+    def __init__(self, node_key, parent, depth, leaf_callback, is_raw=False):
+        """Create a new SyncRequest for a given HexaryTrie node.
+
+        :param node_key: The node's key.
+        :param parent: The node's parent.
+        :param depth: The ndoe's depth in the trie.
+        :param leaf_callback: A callback called when for all leaf children of this node.
+        :param is_raw: If True, HexaryTrieSync will simply store the node's data in the db,
+        without decoding and scheduling requests for children. This is needed to fetch contract
+        code when doing a state sync.
+        """
+        self.node_key = node_key
+        self.parents = []
+        if parent is not None:
+            self.parents = [parent]
+        self.depth = depth
+        self.leaf_callback = leaf_callback
+        self.is_raw = is_raw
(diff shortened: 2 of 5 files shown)


trie/hexary.py
```diff
@@ -171,29 +172,12 @@ def _get_node(self, node_hash):
...
-        return encoded_node_hash
-
-    def _decode_node(self, encoded_node_or_hash):
-        if encoded_node_or_hash == BLANK_NODE:
-            return BLANK_NODE
-        elif isinstance(encoded_node_or_hash, list):
-            return encoded_node_or_hash
-        else:
-            return rlp.decode(encoded_node_or_hash)
+        return persist_node(self.db, node)
```
pipermerriam: Seems like just removing the `_persist_node` method and changing all internal uses of `self._persist_node` would be cleaner.

trie/sync.py
```diff
@@ -0,0 +1,179 @@
...
+        self.db = db
+        self.root_hash = root_hash
+        self.logger = logger
+        self.schedule(root_hash, parent=None, depth=0, leaf_callback=self.leaf_callback)
+
+    def leaf_callback(self, data, parent):
+        """Called when we reach a leaf node.
+
+        Should be implemented by subclasses that need to perform special handling of leaves.
+        """
```
pipermerriam, line 66: Can you add an explicit `pass` statement to this method?  Maybe it's just my personal preference but this *empty* method looks confusing to me.

trie/sync.py
```diff
@@ -0,0 +1,179 @@
...
+        elif node_type == NODE_TYPE_BRANCH:
+            depth = request.depth + 1
+            for item in node[:16]:
+                if is_blank_node(item):
+                    continue
+
+                # In a branch, the first 16 items are either a node whose RLP-encoded
+                # representation is under 32 bytes or a reference to another node.
+                if len(item) == 2:
+                    assert get_node_type(item) == NODE_TYPE_LEAF
```
pipermerriam: *nitpick* 

I'd be more comfortable with explicit exceptions over assertions in productions code since assertions are easy to disable.

trie/sync.py
```diff
@@ -0,0 +1,179 @@
...
+
+    def schedule(self, node_key, parent, depth, leaf_callback, is_raw=False):
+        """Schedule a request for the node with the given key."""
+        request = SyncRequest(node_key, parent, depth, leaf_callback, is_raw)
+        if request.node_key in self.db:
+            self.logger.debug("Node %s already exists in db" % encode_hex(request.node_key))
+            return
+
+        for ancestor in request.parents:
+            ancestor.dependencies += 1
```
pipermerriam, line 85: Any reason to not change this to `ancestor.dependencies += len(request.parents)`

trie/sync.py
```diff
@@ -0,0 +1,179 @@
...
+        """Called when we reach a leaf node.
+
+        Should be implemented by subclasses that need to perform special handling of leaves.
+        """
+
+    def next_batch(self, n=1):
+        """Return the next requests that should be dispatched."""
+        if len(self.queue) == 0:
+            return []
+        batch = self.queue[-n:]
```
pipermerriam: Maybe perform the `list(reversed())` here instead of in the return statement?  I'm not sure if it actually achieves any performance gains, but it seems like it avoids double allocating this value.

pipermerriam (approved):
Code looks good but I'm not entirely grasping the nuances of how this works (and I don't necessarily need to).  Here are my questions, none of which are blockers.

- I'm curious what the memory profile looks like for this.  For the mainnet chain is there an upper bound of how much we'll have sitting stored in memory?  Is it a depth-first approach meaning we'll only every have approximately 1 full trie branch in memory at a time?
- How does this work with starting and resuming a sync?
- Given that this is part of the syncing logic, why did you choose to implement this here rather than in `py-evm`?

trie/hexary.py
gsalgado: I kept it because then callsites don't have to pass `self.db` as the first argument. Don't feel strong about it, though, so could do as you suggest if you do. But just realized that I'm no longer using that function in TrieSync, so I can revert this change.

trie/sync.py
gsalgado: Sure

trie/sync.py
gsalgado: That's not what this code is actually doing here -- it's incrementing the `.dependencies` of each ancestor by 1

gsalgado:
On 9 January 2018 at 01:42, Piper Merriam <notifications@github.com> wrote:
>
> @pipermerriam approved this pull request.
>
> Code looks good but I'm not entirely grasping the nuances of how this works (and I don't necessarily need to). Here are my questions, none of which are blockers.
>
> I'm curious what the memory profile looks like for this. For the mainnet chain is there an upper bound of how much we'll have sitting stored in memory? Is it a depth-first approach meaning we'll only every have approximately 1 full trie branch in memory at a time?

It ran for several hours when syncing the ropsten state and never
seemed to use a significant amount of memory. Indeed it is a
depth-first approach, so yeah, one full trie branch at a time, but we
will also have SyncRequest instances for every sibling of that branch.
That didn't seem to be a problem as those are discarded whenever the
node they refer to is committed.

> How does this work with starting and resuming a sync?

It can resume a sync because we only commit when we have a full trie
branch and the `schedule()` method will not create a SyncRequest if a
node key is already present in our DB.

> Given that this is part of the syncing logic, why did you choose to implement this here rather than in py-evm?

Because it needs access to internal bits and pieces of py-trie,
including writing directly to the `db`

gsalgado:
@pipermerriam Do you want to have another look or is it ok for me to go ahead and merge it?

trie/sync.py
pipermerriam: Ahhh..... I misread.

pipermerriam:
Merge away.

pipermerriam:
Released in `v1.1.0`

PR #28, opened 2018-01-11
Implement a node iterator

gsalgado:
My plan is to use this to validate the state trie we download after a fast-sync.
I've looked at pyethereum for "inspiration" (i.e. lifted most of the code from
there), and have been trying to make it more straightforward/cleaner, but I'm
kinda stuck, so decided to submit it as is

### DIFF
--- a/tests/test_iter.py
+++ b/tests/test_iter.py
@@ -1,0 +1,75 @@
+import json
+import os
+
+import pytest
+
+from hypothesis import (
+    given,
+    settings,
+    strategies,
+)
+
+from trie import HexaryTrie
+from trie.iter import NodeIterator
+from trie.utils.nodes import is_extension_node
+from .utils import make_random_trie
+
+
+ROOT_PROJECT_DIR = os.path.dirname(os.path.dirname(__file__))
+NEXT_PREV_FIXTURE_PATH = os.path.join(
+    ROOT_PROJECT_DIR, 'fixtures', 'TrieTests', 'trietestnextprev.json')
+RAW_NEXT_PREV_FIXTURES = [
+    (os.path.basename(NEXT_PREV_FIXTURE_PATH), json.load(open(NEXT_PREV_FIXTURE_PATH)))
+]
+NEXT_PREV_FIXTURES = [
+    ("{0}:{1}".format(fixture_filename, key), fixtures[key])
+    for fixture_filename, fixtures in RAW_NEXT_PREV_FIXTURES
+    for key in sorted(fixtures.keys())
+]
+
+
+@pytest.mark.parametrize(
+    'fixture_name,fixture', NEXT_PREV_FIXTURES,
+)
+def test_trie_next_prev_using_fixtures(fixture_name, fixture):
+    trie = HexaryTrie(db={})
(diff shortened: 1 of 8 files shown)


tests/test_iter.py
```diff
@@ -0,0 +1,77 @@
...
+def test_trie_next_prev_using_fixtures(fixture_name, fixture):
+    trie = HexaryTrie(db={})
+    for k in fixture['in']:
+        k = force_bytes(k)
+        trie[k] = k
+
+    iterator = NodeIterator(trie)
+    for point, _, nxt in fixture['tests']:
+        point = force_bytes(point)
+        nxt = force_bytes(nxt)
```
pipermerriam: You shouldn't use `force_bytes` anymore.  It's going to be deprecated due to some unfixable encoding issues.

trie/iter.py
```diff
@@ -0,0 +1,93 @@
...
+            return curr_key
+        elif node_type == NODE_TYPE_EXTENSION:
+            curr_key = extract_key(node)
+            sub_node = self.trie._get_node(node[1])
+            return curr_key + self._get_next(sub_node)
+        elif node_type == NODE_TYPE_BRANCH:
+            if node[16]:
+                return (16,)
+            scan_range = range(16)
+            for i in scan_range:
```
pipermerriam: *nitpick*

Odd to me that this isn't just `for i in range(16)`

trie/iter.py
```diff
@@ -0,0 +1,93 @@
...
+    def _get_next(self, node):
+        node_type = get_node_type(node)
+        if node_type == NODE_TYPE_BLANK:
+            return None
+        elif node_type == NODE_TYPE_LEAF:
+            curr_key = extract_key(node)
+            return curr_key
+        elif node_type == NODE_TYPE_EXTENSION:
+            curr_key = extract_key(node)
+            sub_node = self.trie._get_node(node[1])
```
pipermerriam: It'd be nice if we could avoid using the internal `trie._get_node`.  Maybe instead we could add a new method on the trie class for retrieving *raw* nodes or something?  `Trie.get_raw_node`?

pipermerriam (approved):
Looks good.  We can address performance issues later when we actually know what they are.

trie/iter.py
gsalgado: Yep, that was some leftover from a refactoring, will change

trie/iter.py
gsalgado: Well, isn't that just a matter of renaming it to `trie.get_node()` (I'd prefer that over `get_raw_node()` as the latter kind of suggests it'd return the RLP-encoded node)?

tests/test_iter.py
gsalgado: Should I just `.encode('utf-8')` until the new APIs land?

pipermerriam: Yes, I think that is fine.

trie/iter.py
pipermerriam: My mistake.  `trie.get_node()` is fine.  For some reason I thought there was already a `get_node`, hence the new name suggestion.

Issue #32, opened 2018-02-02
The different actions between BinaryTrie and HexaryTrie


hwwhww:
* Version: 1.1.0 or the latest commit (4b78035f0dc1314cbda3c58070ae1fe074fe08a1)
* Python: 3.6
* OS: osx

### What was wrong?
When I was trying to switch the transaction and receipt trie from `HexaryTrie` to `BinaryTrie` for `ShardingVM` (https://github.com/ethereum/py-evm/issues/331), I found that the APIs results of these two tries are slightly different in dealing *not-found-in-dict*.

`HexaryTrie` returns `BLANK_NODE`
https://github.com/ethereum/py-trie/blob/4b78035f0dc1314cbda3c58070ae1fe074fe08a1/trie/hexary.py#L69-L79

`BinaryTrie` returns `None` 
https://github.com/ethereum/py-trie/blob/4b78035f0dc1314cbda3c58070ae1fe074fe08a1/trie/binary.py#L57-L86

#### Code to reproduce the error
```python
from trie import (
    BinaryTrie,
    HexaryTrie,
)


hexary_trie = HexaryTrie(db={})
assert b'hello' not in hexary_trie

binary_trie = BinaryTrie(db={})
assert b'hello' not in binary_trie
```

result: 
```
Traceback (most recent call last):
  File "hello.py", line 16, in <module>
    assert b'hello' not in binary_trie
AssertionError
```
^^^^ That's false.

I replaced `None` with `BLANK_NODE` in `trie._get()` function,  and then the tests in py-evm are passed.

I'm afraid that I may break `BinaryTrie`, so I wanna check with @NIC619, could I simply replace `None` with `BLANK_NODE` in `BinaryTrie._get()` function?

pipermerriam:
Looks like we may want to update `HexaryTrie` to return `None` instead of `BLANK_NODE`?

PR #33, opened 2018-02-03
Fix: BinaryTrie.exist and node override error check

NIC619:
### What was wrong?
Part 1. As pointed out in #32, `b'hello' not in binary_trie` should return `True` .
Part 2. Update in #31 misplace the node override check in `_set_branch_node` and causing some value updates to fail as pointed out in #34 .

### How was it fixed?
Part 1. `BinaryTrie.exist()` should check the returned value against `None` instead of `BLANK_NODE`.
Also remove `BLANK_NODE` and `BLANK_NODE_HASH` which are only related to `HexaryTrie`.
Part 2. Move the node override check to `_set`.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/tests/test_bin_trie.py
+++ b/tests/test_bin_trie.py
@@ -103,12 +103,26 @@ def test_bin_trie_invalid_key(invalide_key, if_error):
 
 
 @given(keys=st.lists(st.binary(min_size=32, max_size=32), min_size=100, max_size=100, unique=True),
-       chosen_numbers=st.lists(st.integers(min_value=0, max_value=99), min_size=50, max_size=50))
+       chosen_numbers=st.lists(
+           st.integers(min_value=0, max_value=99),
+           min_size=50,
+           max_size=50,
+           unique=True)
+       )
 @settings(max_examples=10)
 def test_bin_trie_update_value(keys, chosen_numbers):
+    """
+    This is a basic test to see if updating value works as expected.
+    """
     trie = BinaryTrie(db={})
     for key in keys:
         trie.set(key, b'old')
-    
+
+    current_root = trie.root_hash
     for i in chosen_numbers:
+        trie.set(keys[i], b'old')
+        assert current_root == trie.root_hash
         trie.set(keys[i], b'new')
+        assert current_root != trie.root_hash
+        assert trie.get(keys[i]) == b'new'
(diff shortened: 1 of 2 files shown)


hwwhww:
It works for the trie-switching, but I wonder why we use different logic for `BinaryTrie` and `HexaryTrie` in deep?

NIC619:
We could change `None` to `BLANK_NODE` in `BinaryTrie` for uniformity. But, in my understanding, basically in `BinaryTrie` there's no blank node and two child nodes of every node must be `KV_NODE`, `BRANCH_NODE` or `LEAF_NODE`(with value) while in `HexaryTrie` one node have sixteen child nodes and those that don't have a value are initialized as `BLANK_NODE`.

hwwhww:
This PR patch works for `py-evm` tests, thanks.

tests/test_bin_trie.py
```diff
@@ -100,3 +100,15 @@ def test_bin_trie_invalid_key(invalide_key, if_error):
         previous_root_hash = trie.root_hash
         trie.delete(invalide_key)
         assert previous_root_hash == trie.root_hash
+
+
+@given(keys=st.lists(st.binary(min_size=32, max_size=32), min_size=100, max_size=100, unique=True),
+       chosen_numbers=st.lists(st.integers(min_value=0, max_value=99), min_size=50, max_size=50))
+@settings(max_examples=10)
+def test_bin_trie_update_value(keys, chosen_numbers):
```
pipermerriam: This test has no assertions. I'm confused as to what it's doing.  If there are no reasonable assertions to make, can you update the test with a docstring explaining what it's doing and why there are no assertions.

pipermerriam (changes requested):

tests/test_bin_trie.py
NIC619: Sorry about the confusion. It's a basic test to see if `BinaryTrie.update` is working as expected. Just making sure that basic functionality like `update` works correctly.

Added doctoring for explanation and conditions check.

pipermerriam (approved):

pipermerriam:
I'll get this merged and released when tests are green.

pipermerriam:
Released as `1.2.0`.

Issue #38, opened 2018-03-30
HexaryTrieSync instantiates SyncRequestAlreadyProcessed exception but does not raise it


gsalgado:
https://github.com/ethereum/py-trie/blob/master/trie/sync.py#L164

PR #41, opened 2018-04-03
A couple misc fixes to HexaryTrieSync

gsalgado:
- Actually raise SyncRequestAlreadyProcessed() (#38)
- Avoid unnecessarily re-hashing the node data when committing
- Keep track of committed nodes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,6 @@ matrix:
     - python: "3.5"
       env: TOX_POSARGS="-e flake8"
     # core
-    - python: "3.4"
-      env: TOX_POSARGS="-e py34"
     - python: "3.5"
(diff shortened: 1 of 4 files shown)


gsalgado:
@carver @pipermerriam looks like the latest version of eth-hash no longer supports py3.4. should I just pin the version used here to something older?

pipermerriam:
Yes, we're fine to drop 3.4 support here.

gsalgado:
@pipermerriam updated to drop py3.4 support

pipermerriam (approved):
Is the `committed_nodes` counter just for informational tracking purposes?

gsalgado:
Yep, the `committed_nodes` attribute is used in a py-evm branch I'm working on, when reporting progress of the state sync

PR #43, opened 2018-04-05
Cache keys of existing nodes in HexaryTrieSync to improve performance

gsalgado:


trie/sync.py
```diff
@@ -58,8 +58,12 @@ def __init__(self, root_hash, db, logger):
         self.db = db
         self.root_hash = root_hash
         self.logger = logger
-        self.schedule(root_hash, parent=None, depth=0, leaf_callback=self.leaf_callback)
+        # A cache of node hashes we know to exist in our DB, used to avoid querying the DB
+        # unnecessarily as that's the main bottleneck when dealing with a large DB like for
+        # ethereum's mainnet/ropsten.
+        self._existing_nodes = set()
```
pipermerriam, line 64: What do you think about using either an LRU cache or a bloom filter for this?  Alternatively, can you fill me in on what the upper bounds are for how big this set may grow?

pipermerriam (approved):

trie/sync.py
gsalgado: During the (somewhat limited) testing I did, it didn't seem likely that an LRU cache would provide the same improvements, but I'd need to confirm that.

Although I ran this for a couple hours and it didn't seem to use any more memory than usual, I'm thinking that maybe it'd make more sense to implement this in py-evm, as a caching layer on top of the `ChainDB` that we pass to `StateDownloader`, as (in theory) this code could be used with a DB that is just a plain dictionary and in that case this hack would not provide any benefit. What do you think?

pipermerriam: Generally agree.  Going to link to https://github.com/ethereum/py-evm/issues/507 as again, I think this fits into the realm of what that is trying to fix.  Instead of doing this at this level, in theory we could just apply a thin *wrapper* around the underlying db which cached these existence checks.

Fine to move this forward however you'd like.  If any of these *better* solutions seem promising as down-the-road fixes for this lets open an issue to track that.

pipermerriam:
released along with #45 via `v1.3.4`

PR #44, opened 2018-04-09
improve performance of nibbles helper functions

pipermerriam:
### What was wrong?

The `bytes_to_nibbles` function had some un-necessary overhead.

### How was it fixed?

Removed it by using a more streamlined approach.

#### Cute Animal Picture

9b6d27b2deacffba1d087d7314200e44


pipermerriam:
I suspect this will have *some* measurable effect on trie performance.

gsalgado (approved):
Looks good, but it doesn't seem to have any effect when doing a regular sync

PR #45, opened 2018-04-12
Piper/faster rlp insertions

pipermerriam:
### What was wrong?

The RLP encoding on inserts could be faster.

### How was it fixed?

Used `rlp.codec.encode_raw` which only handles `bytes` and lists of `bytes`.

#### Baseline (current master @ de7b644d306c54f21092d7ed562a952c92940d08)

```
$ python tests/speed.py
testing 15488 values
time to insert 15488 - 11.46
time to read 15488 - 2.59
```
      
#### This branch

```
$ python tests/speed.py
testing 15488 values
time to insert 15488 - 7.05
time to read 15488 - 2.51
```

#### Cute Animal Picture

cute_baby_camel_559707

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,6 @@ fixtures/**
 
 # profiling
 prof/**
+
+# hypothesis
+.hypothesis
--- a/tests/speed.py
+++ b/tests/speed.py
@@ -1,0 +1,38 @@
+import time
+import random
+
+import itertools
+
+from trie import HexaryTrie
+
+
+def mk_random_bytes(n):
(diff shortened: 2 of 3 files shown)


pipermerriam:
@gsalgado cc: decent performance boost here.

gsalgado (approved):
Awesome!

gsalgado:
FWIW, I just profiled the full-sync code with/without this change and whereas before we were spending ~10% of the time on `rlp.encode`, I'm now seeing only ~3% of the time spent on `rlp.encode_raw`, and an insignificant amount of time spent on `rlp.encode`, so massive :+1: for this!

PR #48, opened 2018-04-17
WIP: Immutable tries that return a TrieDelta after every change

carver:
## Work in Progress

exploring immutable tries

would need much better performance to be considered for merge.

#### Cute Animal Picture

Cute animal picture


carver:
I still think this is a good idea, but would require too large of a refactor to py-evm to take advantage of. Will return to later.

PR #49, opened 2018-04-18
Significantly reduce writes to DB

carver:
When using the new `trie.squash_changes()` context, the `HexaryTrie` will automatically prune all database changes that don't make it into the final root.

### What was wrong?

As the trie adds & removes keys, it stores all the state in the database. It also stores all of the state generated inside a single `set()`, when modifying each node that it needs to touch, even when that state is unreachable by the end of the action. This creates a lot of cruft data that is hard to track down, and increases write volume significantly.

### How was it fixed?

Added a new `prune` flag which aggressively removes *all* old node data. To batch up changes, and throw out the data generated along the way, use the new context manager, like:

```py
trie = HexaryTrie(HypotheticalDiskDatabase())

with trie.squash_changes() as squashing_trie:
    squashing_trie[key1] = val1
    squashing_trie[key2] = val2
    del squashing_trie[key3]
```

The two state roots (after `key1` and `key2`), and in fact **all** of the nodes created and unreachable from the latest state root, are never pushed to the database on disk. See `test_hexary_trie_saving_final_root()` which tests that all unreachable fields have been pruned away.

`ScratchDB` was largely lifted from https://github.com/jannikluhn/py-evm/commit/40e46121eaba893997976ebd08fbbf5f3a9e885e -- thanks @sunxivincent

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/tests/test_hexary_trie.py
+++ b/tests/test_hexary_trie.py
@@ -8,11 +8,15 @@ import os
 from eth_utils import (
     is_0x_prefixed,
     decode_hex,
+    keccak,
     text_if_str,
     to_bytes,
 )
 
 from trie import HexaryTrie
+from trie.utils.nodes import (
+    decode_node,
+)
 
 
 def normalize_fixture(fixture):
@@ -61,7 +65,7 @@ RAW_FIXTURES = tuple(
 )
 
 
-FIXTURES = tuple(
+FIXTURES_NORMALIZED = tuple(
     (
         "{0}:{1}".format(fixture_filename, key),
         normalize_fixture(fixtures[key]),
@@ -71,11 +75,11 @@ FIXTURES = tuple(
 )
 
 
-@pytest.mark.parametrize(
-    'fixture_name,fixture', FIXTURES,
-)
-def test_trie_using_fixtures(fixture_name, fixture):
+def get_fixture_iterator(fixture):
+    return itertools.islice(itertools.permutations(fixture['in']), 100)
 
+
+def get_expected_results(fixture):
     keys_and_values = fixture['in']
     deletes = tuple(k for k, v in keys_and_values if v is None)
     remaining = {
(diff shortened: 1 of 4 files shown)


trie/utils/db.py
```diff
@@ -0,0 +1,63 @@
...
+    #
+    # if not key is found, return None
+    def __getitem__(self, key):
+        if key in self.cache:
+            return self.cache[key]
+        else:
+            try:
+                current_value = self.wrapped_db.get(key)
+            except KeyError:
+                current_value = None
```
pipermerriam: I think both the `try` and `except` block might be clearer with their own `return` statements.

trie/utils/db.py
```diff
@@ -0,0 +1,63 @@
...
+            return current_value
+
+    def __setitem__(self, key, value):
+        self.cache[key] = value
+
+    def __delitem__(self, key):
+        val = self.cache.pop(key, MISSING)
+        if val is MISSING:
```
pipermerriam: I don't understand this.  `val == Missing` when the key is not present in the cache.  Why do we only add it to the deletes set if the key was missing?  Seems like it should be added on any delete?  I'm missing something.

trie/utils/db.py
```diff
@@ -0,0 +1,63 @@
...
+    which represents as a dictionary of database keys and values.
+    This class should be usable as a context manager, the changes either all fail or all succeed.
+    Upon exiting the context, it writes all of the key value pairs from the cache into
+    the underlying database. If anyerror occurred before committing phase,
+    we would not apply commits at all.
+    """
+    def __init__(self, wrapped_db):
+        self.wrapped_db = wrapped_db
+        self.cache = {}  # type: Dict[bytes, bytes]
+        self.deletes = []
```
pipermerriam: This seems like it should be a `set`

pipermerriam (commented):
Only partial review.  Will finish tomorrow.

trie/utils/db.py
carver: Either the value was in the cache or the wrapped_db. If the value was added to the cache, then we delete it by removing it from the cache. Otherwise, we queue it up for deletion from the wrapped database at commit time.

tests/test_hexary_trie.py
```diff
@@ -109,3 +121,122 @@ def test_trie_using_fixtures(fixture_name, fixture):
...
+        unread = flagged_usage_db.unread_keys()
+        straggler_data = {k: (db[k], decode_node(db[k])) for k in unread}
+        assert len(unread) == 0, straggler_data
+
+        expected_root = fixture['root']
+        actual_root = trie.root_hash
+        assert actual_root == expected_root
+
+
+def test_hexarie_trie_saves_old_keys_after_batch_save():
```
gsalgado: I'm confused by this test's name as it doesn't seem to assert anywhere that the previous value is still present in the trie. I guess it's just missing that check at the end?

tests/test_hexary_trie.py
```diff
@@ -109,3 +121,122 @@ def test_trie_using_fixtures(fixture_name, fixture):
...
+    old_root_hash = trie.root_hash
+
+    with trie.squash_changes() as memory_trie:
+        memory_trie.set(b'what floats on water?', b'a duck')
+
+    assert trie[b'what floats on water?'] == b'a duck'
+    assert trie[b'what is your favorite color?'] == b'I don\'t know that'
+
+
+def test_hexary_trie_batch_save_keeps_last_root_data():
```
gsalgado: Or maybe this one is enough and we can drop the above test?

trie/hexary.py
```diff
@@ -46,18 +50,18 @@
...
-    root_hash = None
+class HexaryTrie:
+    __slots__ = ('db', 'root_hash', 'is_pruning')
 
     # Shortcuts
     BLANK_NODE_HASH = BLANK_NODE_HASH
     BLANK_NODE = BLANK_NODE
 
-    def __init__(self, db, root_hash=BLANK_NODE_HASH):
+    def __init__(self, db, root_hash=BLANK_NODE_HASH, prune=False):
```
gsalgado, line 60: Just curious if you have any idea what the performance impact of using `prune=True` would be. Do you think it'd make sense for us to use that by default in py-evm or should we just use the `squash_changes()` contextmanager when generating tx/receipt tries?

gsalgado (approved):
I ran a quick fast-sync test using this on `make_trie_root_and_nodes` and it seems to reduce our DB footprint by almost 40%. Great work!

trie/utils/db.py
pipermerriam:
```python
wrapped_db = {b'a': 4}
cached_db = CachedDB(wrapped_db)
cached_db[b'a'] = 3  # b'a' is now in the cache, db[b'a'] -> 3
del cached_db[b'a']  # b'a' is not in the cache, b'a' not in delete set, db[b'a'] -> 4  (wrong!)
```

What am I missing?  It seems like *any* key deletion should be queued for deletion from the wrapped database.

trie/utils/db.py
```diff
@@ -0,0 +1,63 @@
...
+    def __setitem__(self, key, value):
+        self.cache[key] = value
+
+    def __delitem__(self, key):
+        val = self.cache.pop(key, MISSING)
+        if val is MISSING:
+            self.deletes.append(val)
+
+    def __contains__(self, key):
+        return key in self.cache or key in self.wrapped_db
```
pipermerriam: This will fail if a key is in the delete set but still present in the underlying db, returning `True` when it *should* return `False`.

```
if key in self.deletes:
    return False
elif key in cache or key in self.wapped_db:
    return True
```

trie/utils/db.py
```diff
@@ -0,0 +1,63 @@
...
+    #
+    # Dictionary API
+    #
+    # if not key is found, return None
+    def __getitem__(self, key):
+        if key in self.cache:
+            return self.cache[key]
+        else:
+            try:
+                current_value = self.wrapped_db.get(key)
```
pipermerriam: I think this will return incorrectly if the key is in the delete set.  Seems like another check needs to be added to return `None` if the key is in the delete set.

trie/hexary.py
```diff
@@ -391,3 +422,16 @@ def __delitem__(self, key):
...
         return self.exists(key)
+
+    @contextlib.contextmanager
+    def squash_changes(self):
+        scratch_db = ScratchDB(self.db)
+        try:
+            with scratch_db.batch_commit(do_deletes=self.is_pruning):
+                memory_trie = type(self)(scratch_db, self.root_hash, prune=True)
+                yield memory_trie
+        except Exception as exc:
```
pipermerriam: I think that this whole `try/except/else` block can just be removed.  If an exception is raised, the statement which updates the `root_node` just won't be executed.

pipermerriam (approved):
I think the scheme of storing `deletes` separately is making things more difficult.  I think that using `None` in your cache to represent a deleted key will clean up your logic and make things simpler as well as removing the possibility of having conflicting values with a key present in the cache and the delete set.

tests/test_hexary_trie.py
```diff
@@ -84,9 +88,17 @@ def test_trie_using_fixtures(fixture_name, fixture):
...
+@pytest.mark.parametrize(
+    'fixture_name,fixture', FIXTURES,
+)
+def test_trie_using_fixtures(fixture_name, fixture):
 
-    for kv_permutation in itertools.islice(itertools.permutations(keys_and_values), 100):
-        print("in it")
+    remaining, deletes = get_expected_results(fixture)
+
+    for kv_permutation in get_fixture_iterator(fixture):
```
pipermerriam: Thoughts on using a parameterized fixture instead of this loop so that we get a clearer definition of which permutation(s) fail?

tests/test_hexary_trie.py
```diff
@@ -109,3 +121,122 @@ def test_trie_using_fixtures(fixture_name, fixture):
...
+        super().__init__(*args, **kwargs)
+        self.read_keys = set()
+
+    def __getitem__(self, key):
+        result = super().__getitem__(key)
+        self.read_keys.add(key)
+        return result
+
+    def unread_keys(self):
+        return set(self) - self.read_keys
```
pipermerriam: Am I the only one that gets curmudgeonly when it comes to pythons default to iterate over just keys in dictionaries?

tests/test_hexary_trie.py
```diff
@@ -109,3 +121,122 @@ def test_trie_using_fixtures(fixture_name, fixture):
...
+        actual_root = trie.root_hash
+        assert actual_root == expected_root
+
+
+def test_hexarie_trie_saves_old_keys_after_batch_save():
+    db = {}
+    trie = HexaryTrie(db)
+    trie.set(b'what floats on water?', b'very small rocks')
+    trie.set(b'what is your favorite color?', b'I don\'t know that')
+    old_root_hash = trie.root_hash
```
pipermerriam: Looks like you store this but never use it?

trie/hexary.py
pipermerriam: Seems like pruning on by default would be a good thing.

trie/utils/db.py
carver: Because the db of a trie always stores `db[keccak(val)] = val`, having different key/values in the wrapped and cached db would mean finding two different pre-images with the same sha3 hash. I don't know if there's only one right thing to do in that situation, but leaving the old one there when pruning the new one doesn't seem insane (the only cost being a little extra database usage). I also don't see anything incorrect about deleting always, but it causes a performance hit:

1. We carry around a lot more deletes to try to run on the wrapped db
2. We would now have to test if the key exists before every delete operation on `wrapped_db`

I could go either way, and I don't think it's hard to change later. Thoughts?

tests/test_hexary_trie.py
carver: Yeah, on another review, I think we cover everything in the one below. I'll remove this one.

trie/hexary.py
carver: It's a pretty significant change, since it means you can no longer look up state at old root hashes by default. It would be a major version bump kind of change, since anyone already using it would get hit with pretty nasty bugs if they expect old state to stick around by default.

I'm not opposed, other than the backward compatibility concern.

I was personally imagining using the context manager so that we prune intermediate states, but still can look up old states. In fact, we probably *have* to do this to deal with chain re-orgs.

trie/utils/db.py
pipermerriam: So ignoring the hash collision which I think is safe to ignore, I think it's still worth looking at this case.

```
wrapped_db = {b'a': 3}
cached_db = CachedDB(wrapped_db)
cached_db[b'a'] = 3  # b'a' is now in the cache, db[b'a'] -> 3  SAME AS BEFORE
del cached_db[b'a']  # b'a' is not in the cache, b'a' not in delete set, db[b'a'] -> 3
```

I think this is what you were mentioning about taking a performance hit to have to hit the underlying db with every delete.  Still, this seems like something we should do and we can revisit if the performance hit worth addressing.

trie/hexary.py
pipermerriam: Fair, and I think we do actually make this assumption of being able to roll back to previous state roots in py-evm.  

I'm still inclined to make this the default behavior as the current behavior is not really documented and our use of it is largely *implicit* based on the implementation.  Maybe it's worth doing some minor experimentation after releasing this with it off by default to see how bad it changes things by turning it on by default within py-evm.

tests/test_hexary_trie.py
carver: I'm not opposed. Any failure ends up printing kv_permutation, and this whole batch only takes ~1s to run, so it wasn't a major barrier to getting work done. But a test for each permutation is more stylistically consistent. :balance_scale: 

I guess I'll switch it to a test per permutation.

tests/test_hexary_trie.py
carver: Haha, I suppose at first it bothered me. I expected `.items()` to be the default. Now it's just like little quirks in the English language: I know they're terrible, but I don't think about them anymore.

I'm happy to be explicit here and use `self.keys()`, though.

pipermerriam: I was more just voicing my grumblings.  Don't really care, this is a test.

PR #54, opened 2018-04-26
fixes #53

### LINKED ISSUE #53: Non-deterministic test failure in trie sync.
* Version: 1.3.5
* Python: 2.7/3.4/3.5
* OS: osx/linux/win

### What was wrong?

The sync tests may have a non-deterministic *something* that resulting in a CI failure.

https://gist.github.com/pipermerriam/bf3a360b896f75be39b363495821a8aa

Not to fail, or to fail reliably.

pipermerriam:
### What was wrong?

The `HexaryTrieSync` was coded primarily to support the primary use case, syncing the ethereum state trie.  Because of this, it was written to not support certain key/value types that won't appear in the ethereum state trie because of the way that keys and values are encoded.

however, our fuzz testing ran into this case and caused CI failures.

### How was it fixed?

Went ahead and added support for nodes who's RLP encoding is less than 32 bytes and thus, rather than being referenced, they are actually inlined.

#### Cute Animal Picture

cute-camel

### DIFF
--- a/tests/test_sync.py
+++ b/tests/test_sync.py
@@ -16,10 +16,19 @@ from .utils import make_random_trie
 logger = logging.getLogger()
 
 
+# produces a branch node with an extention node who's encoding is less than 32
+# bytes in length so it is inlined.
+EXAMPLE_37968 = 37968
+
+# produces an top level extension node who's encoding is less than 32 bytes in
+# length so it gets inlined.
+EXAMPLE_809368 = 809368
+
+
 @given(random=strategies.randoms())
 @settings(max_examples=50)
-@example(random=RandomWithSeed(37968))
-@example(random=RandomWithSeed(809368))
+@example(random=RandomWithSeed(EXAMPLE_37968))
+@example(random=RandomWithSeed(EXAMPLE_809368))
 def test_trie_sync(random):
     src_trie, contents = make_random_trie(random)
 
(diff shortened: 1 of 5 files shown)


tests/test_sync.py
```diff
@@ -15,7 +17,9 @@
 
 
 @given(random=strategies.randoms())
-@settings(max_examples=10)
+@settings(max_examples=50)
+@example(random=RandomWithSeed(37968))
+@example(random=RandomWithSeed(809368))
```
gsalgado: Maybe add a comment explaining why we want those specific examples?

tests/test_sync.py
```diff
@@ -25,7 +29,11 @@ def test_trie_sync(random):
     while len(requests) > 0:
         results = []
         for request in requests:
-            results.append([request.node_key, src_trie.db[request.node_key]])
+            if len(request.node_key) == 32:
+                results.append([request.node_key, src_trie.db[request.node_key]])
+            else:
+                import rlp
```
gsalgado: Why import rlp here and not as a global?

tests/test_sync.py
```diff
@@ -25,7 +29,11 @@ def test_trie_sync(random):
...
+                results.append([request.node_key, rlp.decode(request.node_key)])
```
gsalgado: Maybe a comment explaining that when the RLP representation of a node is less than 32 bytes that is stored directly as the node's key and that's why we don't look it up in the trie here?

gsalgado (approved):

tests/test_sync.py
pipermerriam: This is accidentally left in.  This code path should no longer happen, removing.

tests/test_sync.py
pipermerriam: accidentally left in.  code path no longer gets traversed, removing.

PR #57, opened 2018-05-27
Use eth_hash.auto for keccak instead of eth_utils

### LINKED ISSUE #55: Use eth-hash.keccak instead of eth-utils.keccak
### What was wrong?

`eth_utils.keccak` enables hashes of multiple types, including text strings and hex strings. This comes at a small performance cost, but py-trie uses keccak heavily so is fairly sensitive to these costs. py-trie only ever submits `bytes` for hashing, so the performance penalty is unnecessary.

### How to fix it

- Replace all usages of `eth_utils.keccak` with `eth_hash.auto.keccak`.
- Add eth-hash as a dependency, and eth-hash[pycryptodome] as a testing dependency.

stefanmendoza:
### What was wrong?

Fixes #55 

### How was it fixed?

Switch from the `eth_utils` keccak function to the `eth_hash.auto`  keccak function.

**NOTE:**
Due to this issue - ethereum/eth-utils#74 - I've bumped the minimum version for eth-utils to 1.0.1[1] and the minimum for eth-hash to 0.1.0 to ensure the dependencies are aligned.

[1] ethereum/eth-utils@dbc4e7c

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -2,4 +2,4 @@ tox==2.6.0
 pytest==3.0.7
 hypothesis==3.7.0
 bumpversion==0.5.3
-eth-hash[pycryptodome]>=0.1.0a4,<1.0.0
+eth-hash[pycryptodome]>=0.1.0,<1.0.0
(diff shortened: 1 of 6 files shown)


stefanmendoza:
@carver ready for review!

pipermerriam (approved):

Issue #37, opened 2018-03-24
Update binary trie with new data structure


NIC619:
As proposed and discussed in [ethresearch](https://ethresear.ch/t/data-availability-proof-friendly-state-tree-transitions/1453/6), Binary Trie built on Sparse Merkle Tree has the advantage that it's more simple and straight forward(there only exist one node type - branch node, just like plain Binary Tree) while size of merkle proof is roughly the same(or even slightly better) since most subtree in Sparse Merkle Tree would be empty and we can compress these nodes into 1 bit representation.

The cost is that it takes O(log(N)) time to `get`/`set` value. In our case(with 2^160 address space) it's 160 nodes to traverse to get to the leaf node, but there should be room for optimization.

fubuloubu:
@NIC619 check out #77

PR #58, opened 2018-07-16
Sparse merkle tree part1

NIC619:
### What was wrong?
Try out Sparse Merkle Tree for sharding research.

### How was it fixed?
Implement Sparse Merkle Tree as described in #37.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/trie/constants.py
+++ b/trie/constants.py
@@ -34,3 +34,10 @@ LEAF_TYPE_PREFIX = bytes([2])
 
 BYTE_1 = bytes([1])
 BYTE_0 = bytes([0])
+
+# Constants for Sparse Merkle Tree
+from eth_hash.auto import keccak
+EMPTY_LEAF_NODE_HASH = BLANK_HASH
+EMPTY_NODE_HASHES = [EMPTY_LEAF_NODE_HASH]
+for _ in range(159):
+    EMPTY_NODE_HASHES.insert(0, keccak(EMPTY_NODE_HASHES[0] + EMPTY_NODE_HASHES[0]))
--- a/trie/constants.py
+++ b/trie/constants.py
@@ -1,3 +1,5 @@
+from eth_hash.auto import keccak
+
 BLANK_NODE = b''
 # keccak(b'')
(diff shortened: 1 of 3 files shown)


trie/constants.py
```diff
@@ -34,3 +34,10 @@
 
 BYTE_1 = bytes([1])
 BYTE_0 = bytes([0])
+
+# Constants for Sparse Merkle Tree
+from eth_hash.auto import keccak
+EMPTY_LEAF_NODE_HASH = BLANK_HASH
+EMPTY_NODE_HASHES = [EMPTY_LEAF_NODE_HASH]
+for _ in range(159):
```
jannikluhn: As 160 (and 159) is used a couple of times, it would make sense to make it a constant. Maybe `NODE_BIT_LENGTH`?

trie/sparse_merkle_tree.py
```diff
@@ -0,0 +1,108 @@
...
+        node_hash = self.root_hash
+        for i in range(160):
+            if path & target_bit:
+                node_hash = self.db[node_hash][32:]
+            else:
+                node_hash = self.db[node_hash][:32]
+            target_bit >>= 1
+
+        if self.db[node_hash] is b'':
+            return None
```
pipermerriam: Reason for returning `None` instead of raising a `KeyError`?  We've found that return value checking is way easier to forget than exception handling.

trie/constants.py
NIC619: Agree, what do you think about maybe something more straight forward like `TREE_HEIGHT`?

jannikluhn: `TREE_HEIGHT` sounds good

trie/sparse_merkle_tree.py
NIC619: No particular reason. Just that in hexary trie, `BLANK_NODE` is returned for non-existent key. Raising an exception does make more sense.

NIC619:
@pipermerriam Updated the behavior for handling non-existent key and ready for review again.

PR #59, opened 2018-07-31
wip: convert HexaryTrieSync methods to async

gsalgado:


carver:
Instead of breaking any current users of the class, maybe leave the old sync version in place and add a new async version of the class? Then it can just be a minor version bump instead of a major version bump.

pipermerriam:
`trie.async.AsyncHexaryTrie`?  probably a nice way to unify the underlying functionality so that we aren't duplicating it.

gsalgado:
I couldn't see a way to do that without duplicating most of the code, and I don't believe we have any other users apart from the StateDownloader, which is why I think we should consider moving this to py-evm: https://github.com/ethereum/py-evm/pull/1124

pipermerriam:
No objections to moving to `py-evm/trinity` codebase.  :+1:

carver:
:+1: for:

1. copying to trinity code base
2. deprecating the class
3. removing in next major version bump

Issue #70, opened 2018-10-17
Update to use `eth_utils.toolz` indirection for imports


pipermerriam:
### What is wrong?

This library currently specifies `cytoolz` as a hard dependency.  This makes pypy support more difficult.

### How can it be fixed?

- Remove `cytoolz` from the direct dependencies.
- Update the `eth-utils` requirement to be `>=1.3.0,<2`
- Update all imports that come from `cytoolz` to instead come from `eth_utils.toolz`

6ug:
I am on it!

Bhargavasomu:
@pipermerriam , as I have not seen any progress on this issue, I have worked on this and fixed this. Please review #76 for merging.

Bhargavasomu:
Also eth-utils 1.3.0 is in beta phase, and have added the same name in the setup.

Issue #69, opened 2018-10-17
Add type hints and expose via PEP561


pipermerriam:
> Largely copy/pasted from https://github.com/ethereum/py-evm/issues/1398

### Background

[Type hints](https://www.python.org/dev/peps/pep-0484/) allow us to perform static type checking, among other things. They raise the security bar by catching bugs at development time that, without type support, may turn into runtime bugs.

[This stackoverflow answer](https://stackoverflow.com/questions/32557920/what-are-type-hints-in-python-3-5/32558710#32558710) does a great job at describing their main benefits.

### What is wrong?

This library currently does not have any type hints.

This needs to be fixed by:

1. Add the `eth-typing>=2.0.0,<3` library as a dependency.
2. Adding all missing type hints.
3. Enforcing (stricter) type checking in CI

### How

There does exist tooling ([monkeytype](https://engineering.instagram.com/let-your-code-type-hint-itself-introducing-open-source-monkeytype-a855c7284881)) to the generation of type hints for existing code bases. From my personal experience `monkeytype` *can* be helpful but does still require manual fine tuning. Also, manually adding these type hints does serve as a great boost to the general understanding of the code base as it forces one to think about the code.

1. Run `mypy --follow-imports=silent --warn-unused-ignores --ignore-missing-imports --no-strict-optional --check-untyped-defs --disallow-incomplete-defs --disallow-untyped-defs --disallow-any-generics -p trie`

2. Eliminate every reported error by adding the right type hint

Because this library supports older versions of python, the type hints will not be able to use the modern python3.6 syntax.

### Definition of done

This issue is done when the following criteria are met:

1. `mypy` is run in CI

Add a new command to the `flake8` environment in the `tox.ini` file that runs:

```
mypy --follow-imports=silent --warn-unused-ignores --ignore-missing-imports --no-strict-optional --check-untyped-defs --disallow-incomplete-defs --disallow-untyped-defs --disallow-any-generics -p py_ecc`
```

2. Usage of `type: ignore` (silencing the type checker) is minimized and there's a reasonable explanation for its usage 

* Update the `setup.py` to expose this data to other libraries as was done here: https://github.com/ethereum/eth-typing/pull/10

### Stretch goals

When this issue is done, stretch goals can be applied (and individually get funded) to tighten type support to qualify:

1. `mypy --strict --follow-imports=silent --ignore-missing-imports --no-strict-optional -p trie`

6ug:
[...]

PR #71, opened 2018-10-19
Removed cytoolz hard dependency and CI fixes

6ug:
### What was wrong?
Fixed https://github.com/ethereum/py-trie/issues/70 with https://github.com/ethereum/py-trie/pull/71/commits/5cd74f8f1435d5408e8f04285701de243eb6f257
- [x] Remove cytoolz from the direct dependencies.
- [x] Update the eth-utils requirement to be >=1.3.0,<2
- [x] Update all imports that come from cytoolz to instead come from eth_utils.toolz

Fixes https://github.com/ethereum/py-trie/issues/69 with https://github.com/ethereum/py-trie/pull/71/commits/a236489d3bd93c3f36abb2e11ce6e296c8f73313
- [x] Add the `eth-typing>=2.0.0,<3` library as a dependency.
- [x] Add a new command to the `flake8` environment in the `tox.ini` file that runs
- [x] Update the `setup.py` to expose data to other libraries
- [ ] Add all missing type hints.

#### Cute Animal Picture

Cute animal picture


pipermerriam:
@6ug I believe you can remove the `.travis.yml` file as it is no longer needed.

pipermerriam:
@6ug looks like you can lower the `eth-utils` to `>=1.2.2`.  `1.3.0` is still in beta.

pipermerriam:
some sort of dependency failure that I'll dig into on Monday if it isn't resolved by then.

I'm also good with dropping support for `rlp<1` and requiring the newer version of the library.

carver:
At least one issue is that mypy was never added as a dependency in setup.py

6ug:
> I'm also good with dropping support for `rlp<1` and requiring the newer version of the library.

@pipermerriam are you talking about dropping `rlp<1` and using newer versions for `rlp<1` ? or second sentence was about libraries in general?

do you know exact supportive versions? I tried with `rlp0: rlp>=1.0.3,<2` and `rlp1: rlp>=1,<2`, it did not worked for me.

pipermerriam:
@6ug 

We currently support two major version ranges of `pyrlp`.  The `0.x` line and the `1.x` line.  I'm proposing that we drop support for `rlp<1` which would allow removal of the `rlp0/rlp1` deliniation in the `tox.ini` and ci runs, and only test against `rlp>=1,<2`.

PR #72, opened 2018-10-19
Modernize

### LINKED ISSUE #68: Modernize project structure
This project was made without the use of our standard project template.

https://github.com/carver/ethereum-python-project-template

The following should be cleaned up to closer match our *modern* project structure.

- remove `requirements-dev.txt` in favor of using `extras_require` pattern.
- add `classifiers` to `setup.py`
- update the readme to contain the appropriate badgest, title and general structure.
- update the `tox.ini` `flake8` environment to use the more modern `lint` name as well as the same approach, installing from `extras_require`.
- add `python_requires` (and exclude `python==3.5.2`)

Extra credit for finding other tangible differences and fixing them.

aashutoshrathi:
Fixes #68 

### What was wrong?
The package needed to be restructured.

#### Cute Animal Picture

980x

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,15 +1,12 @@
 # Python Implementation of the Ethereum Trie structure
 
-
-```shell
-$ pip install trie
-```
+[![PyPI](https://img.shields.io/pypi/v/trie.svg)](https://pypi.org/project/trie/)
+![PyPI - Python Version](https://img.shields.io/pypi/pyversions/trie.svg)
 
 > Warning: This is an early release and is likely to contain bugs as well as
 > breaking API changes.
-
-
-> This library and repository was previously located at https://github.com/pipermerriam/py-trie.  It was transferred to the Ethereum foundation github in November 2017 and renamed to `py-trie`.
+> This library and repository was previously located at [pipermerriam/py-trie](https://github.com/pipermerriam/py-trie). It was transferred to the Ethereum foundation GitHub in 
+> November 2017 and renamed to `py-trie`.
 
 ## Installation
 
@@ -20,9 +17,8 @@ pip install trie
 ## Development
 
 ```sh
-pip install -e . -r requirements-dev.txt
+pip install -e .[dev]
 ```
-
 
 ### Running the tests
 
@@ -34,7 +30,6 @@ py.test tests
 
 Or you can install `tox` to run the full test suite.
 
-
 ### Releasing
 
 Pandoc is required for transforming the markdown README to the proper format to
@@ -42,7 +37,7 @@ render correctly on pypi.
 
 For Debian-like systems:
 
-```
+```sh
 apt install pandoc
 ```
 
@@ -60,7 +55,6 @@ git push && git push --tags
 make release
 ```
 
-
 #### How to bumpversion
 
 The version format for this repo is `{major}.{minor}.{patch}` for stable, and
@@ -73,7 +67,6 @@ If you are in a beta version, `bumpversion stage` will switch to a stable.
 
 To issue an unstable version when the current version is stable, specify the
 new version explicitly, like `bumpversion --new-version 4.0.0-alpha.1 devnum`
-
 
 ## Usage
 
@@ -116,7 +109,8 @@ False
 ```
 
 ## BinaryTrie
-- Note: One drawback of Binary Trie is that **one key can not be the prefix of another key**. For example,
+
+**Note:** One drawback of Binary Trie is that **one key can not be the prefix of another key**. For example,
 if you already set the value `value1` with key `key1`, you can not set another value with key `key` or `key11`
 and the like.
 
@@ -153,7 +147,6 @@ Now Trie looks like this:
             |                           |
   (leafnode, b'value1')       (leafnode, b'value2')
 ```
-
 
 ```python
 >>> # check_if_branch_exist function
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -15,12 +15,21 @@
     include_package_data=True,
     py_modules=['trie'],
     setup_requires=['setuptools-markdown'],
+    python_requires='>=2.7, !=3.5.2',
```
pipermerriam: Lower bound should be `>=3.5` and an upper bound of  `<4` should also be included.

setup.py
```diff
@@ -15,12 +15,21 @@
...
     install_requires=[
         "eth-hash>=0.1.0,<1.0.0",
         "eth-utils>=1.0.1,<2.0.0",
         "rlp>=0.4.7,<2.0.0",
         "cytoolz>=0.8.0,<1",
     ],
+    extras_require=[
+        "tox==2.6.0",
```
pipermerriam: Take a look here for how extras require should be structured:

https://github.com/ethereum/ethereum-python-project-template/blob/master/setup.py#L8-L29

pipermerriam (changes requested):
Can you also remove the legacy `.travis.yml` file.

setup.py
carver: Should be `>=3.5.3` to avoid with some early py3.5 bugs

aashutoshrathi:
@pipermerriam Please trigger build once.
Tests were passing before squash.

aashutoshrathi:
@pipermerriam I think we are good to go.
CircleCI-pyp3 throws some error at their end, some internet problem on their side, check log [here](https://circleci.com/gh/ethereum/py-trie/554?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

pipermerriam:
re-running failed job.

pipermerriam (approved):

pipermerriam:
Looks like CI builds are not triggering (maybe due to the github outage over the weekend).  Will wait to see what happens.

carver:
> Looks like CI builds are not triggering (maybe due to the github outage over the weekend).

Yup, from 9 minutes ago on status.github.com:
> We have temporarily paused delivery of webhooks while we address an issue.

aashutoshrathi:
Everything is normal now :tada:

README.md
```diff
@@ -1,15 +1,11 @@
...
-```
+[![PyPI](https://img.shields.io/pypi/v/trie.svg)](https://pypi.org/project/trie/)
+![PyPI - Python Version](https://img.shields.io/pypi/pyversions/trie.svg)
 
 > Warning: This is an early release and is likely to contain bugs as well as
 > breaking API changes.
-
-
-> This library and repository was previously located at https://github.com/pipermerriam/py-trie.  It was transferred to the Ethereum foundation github in November 2017 and renamed to `py-trie`.
+> This library and repository was previously located at [py-trie](https://github.com/pipermerriam/py-trie).  It was transferred to the Ethereum foundation github in November 2017 and renamed to `py-trie`.
```
carver: I think the old link was better because it highlights that it was hosted under the user pipermerriam. It won't be long before this note can simply be removed.

tox.ini
```diff
@@ -4,21 +4,21 @@ envlist=
...
   py.test {posargs:tests}
 deps =
-  -r{toxinidir}/requirements-dev.txt
   rlp0: rlp>=0.4.7,<1
   rlp1: rlp>=1,<2
 basepython =
   py35: python3.5
   py36: python3.6
   pypy3: pypy3
+extras = test, dev
```
carver: Only developers should need to use the `dev` extra. If tox fails with only the `test` extra, you can move any missing dependencies from dev to test.

carver (changes requested):
The url change is just a suggestion, but I think using the dev extra in tox feels wrong.

README.md
aashutoshrathi: So should I remove the note?
The older one had an MD linting error that is bare URLs are not allowed.

carver: Oh interesting. Maybe change the url text to `pipermerriam/py-trie`

aashutoshrathi: Sure

setup.py
```diff
@@ -15,12 +40,14 @@
     include_package_data=True,
     py_modules=['trie'],
     setup_requires=['setuptools-markdown'],
+    python_requires='>=3.5, !=3.5.2, <4',
```
carver: Why add 3.5.1 as an option again? On mobile, so can't find the link, but I think there was some reason it's not preferable. I don't think think it's a blocker, just curious.

carver (approved):
Cool, thanks, changes look good. Feel free to squash your commits, or we can do it on merge.

setup.py
aashutoshrathi: Not sure the issue mentioned not to allow `3.5.2` only. :sweat_smile:

aashutoshrathi:
That CircleCI :laughing: 
@carver @pipermerriam It fails in pulling changes. Please check it.

carver:
This time it was GitHub that returned a bad result when trying to merge against master. But... that happens from time to time. Our most modern circle scripts reattempt the merge three times in a row. I see that it want explicitly listed, but upgrading to that approach is reasonable to add to this PR, since it's defined in the template project. Check out how the template project does it in the .circleci folder.

pipermerriam:
Good to merge when green.

Issue #65, opened 2018-09-14
Implement Merkle Proof Generation and Validation API


fubuloubu:
For working on things like Plasma and Ethereum 2.0, the Merkle Tree data structure is very critical, especially the generation and validation of merkle proofs. It would be beneficial to implement an API to make this easy where generation takes a key and makes a witness (for the current value in the tree), and validation takes a key, value, and a specific witness and validates that the provided data and witness are valid with the current root of the tree.

The validation part might be done as a separate utility function that could obtain the merkle tree root for that proof and specific key/value pair, and then validates that the root hash matches. This would allow others to validate proofs generated against prior values of the root hash for things like Plasma that need to work with validation of proofs against prior tree roots.

fubuloubu:
Fixed with #77

PR #77, opened 2018-11-19
Sparse Merkle Tree and helpers

fubuloubu:
### What was wrong?
I wanted #58 to have more features that are useful for things like Plasma, like the ability to return Merkle branches, the ability to set key size/tree height, and to return node updates from setting a key so that you could use them to broadcast updates to other clients (see: [merkle sync](https://github.com/GunClear/merkle-sync/))

Solves #65

#### Cute Animal Picture
baby walrus

### DIFF
--- a/tests/test_constants.py
+++ b/tests/test_constants.py
@@ -10,7 +10,30 @@ from trie.constants import (
         BLANK_NODE_HASH,
     )
 
+from trie.smt import SparseMerkleTree as SMT
+
 
 def test_hash_constants():
     assert BLANK_HASH == keccak(BLANK_NODE)
     assert BLANK_NODE_HASH == keccak(encode_raw(b''))
+
+
+def test_smt256_empty_hashes():
+    DEPTH=256  # Default depth is 32 bytes
+
+    # Start at the bottom
+    EMPTY_LEAF_NODE_HASH = BLANK_HASH
+    EMPTY_NODE_HASHES = [EMPTY_LEAF_NODE_HASH]
+
+    # More hashes the lower you go down the tree (to the root)
+    # NOTE: Did this with different code as a sanity check
+    for _ in range(DEPTH - 1):
+        EMPTY_NODE_HASHES.insert(0, keccak(EMPTY_NODE_HASHES[0] + EMPTY_NODE_HASHES[0]))
+    EMPTY_ROOT_HASH = keccak(EMPTY_NODE_HASHES[0] + EMPTY_NODE_HASHES[0])
+
+    smt = SMT()
+    assert smt.root_hash == EMPTY_ROOT_HASH
+
+    key = b"\x00" * 32
+    # _get(key) returns value, branch tuple
+    assert smt._get(key)[1] == EMPTY_NODE_HASHES
--- a/tests/test_constants.py
+++ b/tests/test_constants.py
@@ -1,0 +1,16 @@
+from eth_utils import (
+        keccak,
+    )
+
+from rlp.codec import encode_raw
+
+from trie.constants import (
+        BLANK_NODE,
+        BLANK_HASH,
+        BLANK_NODE_HASH,
+    )
+
+
+def test_hash_constants():
+    assert BLANK_HASH == keccak(BLANK_NODE)
+    assert BLANK_NODE_HASH == keccak(encode_raw(b''))
--- a/tests/test_constants.py
+++ b/tests/test_constants.py
@@ -1,3 +1,5 @@
+import collections
+
 from eth_utils import (
         keccak,
     )
@@ -19,16 +21,16 @@ def test_hash_constants():
 
 
 def test_smt256_empty_hashes():
-    DEPTH=256  # Default depth is 32 bytes
+    DEPTH = 256  # Default depth is 32 bytes
 
     # Start at the bottom
     EMPTY_LEAF_NODE_HASH = BLANK_HASH
-    EMPTY_NODE_HASHES = [EMPTY_LEAF_NODE_HASH]
+    EMPTY_NODE_HASHES = collections.deque([EMPTY_LEAF_NODE_HASH])
 
     # More hashes the lower you go down the tree (to the root)
     # NOTE: Did this with different code as a sanity check
     for _ in range(DEPTH - 1):
-        EMPTY_NODE_HASHES.insert(0, keccak(EMPTY_NODE_HASHES[0] + EMPTY_NODE_HASHES[0]))
+        EMPTY_NODE_HASHES.appendleft(keccak(EMPTY_NODE_HASHES[0] + EMPTY_NODE_HASHES[0]))
     EMPTY_ROOT_HASH = keccak(EMPTY_NODE_HASHES[0] + EMPTY_NODE_HASHES[0])
 
     smt = SMT()
@@ -36,4 +38,4 @@ def test_smt256_empty_hashes():
 
     key = b"\x00" * 32
     # _get(key) returns value, branch tuple
-    assert smt._get(key)[1] == EMPTY_NODE_HASHES
+    assert smt._get(key)[1] == tuple(EMPTY_NODE_HASHES)
--- a/tests/test_smt.py
+++ b/tests/test_smt.py
@@ -17,6 +17,7 @@ from trie.smt import (
 )
 def test_simple_kv(k, v):
     smt = SparseMerkleTree(key_size=len(k))
+    empty_root = smt.root_hash
 
     # Nothing has been added yet
     assert not smt.exists(k)
@@ -24,12 +25,13 @@ def test_simple_kv(k, v):
     # Now that something is added, it should be consistent
     smt.set(k, v)
     assert smt.get(k) == v
+    assert smt.root_hash != empty_root
     assert smt.root_hash == calc_root(k, v, smt.branch(k))
 
     # If you delete it, it goes away
     smt.delete(k)
     assert not smt.exists(k)
-
+    assert smt.root_hash == empty_root
 
 @given(
     data=st.data(),
@@ -71,7 +73,7 @@ def test_branch_updates(data, key_size):
 
         # Merge all of the updates into the tracked proof entries
         for u in proof_updates:
-            p.merge(*u)
+            p.update(*u)
 
-        # All merges should be consistent with the latest smt root
+        # All updates should be consistent with the latest smt root
         assert p.root_hash == smt.root_hash
--- a/tests/test_smt.py
+++ b/tests/test_smt.py
@@ -1,0 +1,34 @@
+import pytest
+
+from eth_utils import (
+        keccak,
+        to_int,
+    )
+
+from trie.smt import (
+    SparseMerkleTree,
+    calc_root,
+)
+
+
+@pytest.mark.parametrize(
+        "k,v",
+        (
+            (b'\x01', b'\x01'),
+            (b'\xff', b'\xff'),
+        )
+    )
+def test_simple_kv(k, v):
+    smt = SparseMerkleTree(keysize=1)
+    
+    # Nothing has been added yet
+    assert not smt.exists(k)
+    
+    # Now that something is added, it should be consistent
+    smt.set(k, v)
+    assert smt.get(k) == v
+    assert smt.root_hash == calc_root(k, v, smt.branch(k))
+    
+    # If you delete it, it goes away
+    smt.delete(k)
+    assert not smt.exists(k)
--- a/tests/test_smt.py
+++ b/tests/test_smt.py
@@ -35,15 +35,15 @@ def test_simple_kv(k, v):
 
 
 @given(
-    key_size = st.shared(st.integers(min_value=1, max_value=32), key="key_size"),
+    key_size=st.shared(st.integers(min_value=1, max_value=32), key="key_size"),
     # Do this so that the size of the keys (in bytes) matches the key_size for the test
-    keys = st.shared(st.integers(), key="key_size").flatmap(lambda key_size: st.lists(
+    keys=st.shared(st.integers(), key="key_size").flatmap(lambda key_size: st.lists(
         elements=st.binary(min_size=key_size, max_size=key_size),
         min_size=3,
         max_size=3,
         unique=True,
     )),
-    vals = st.lists(
+    vals=st.lists(
         elements=st.binary(min_size=1, max_size=32),
         min_size=3,
         max_size=3,
--- a/tests/test_smt.py
+++ b/tests/test_smt.py
@@ -1,34 +1,77 @@
-import pytest
-
-from eth_utils import (
-        keccak,
-        to_int,
+from hypothesis import (
+        given,
+        strategies as st,
     )
 
 from trie.smt import (
     SparseMerkleTree,
+    SparseMerkleProof,
     calc_root,
+    BLANK_NODE,
 )
 
 
-@pytest.mark.parametrize(
-        "k,v",
-        (
-            (b'\x01', b'\x01'),
-            (b'\xff', b'\xff'),
-        )
-    )
+@given(
+    k=st.binary(min_size=1, max_size=32),
+    v=st.binary(min_size=1, max_size=32),
+)
 def test_simple_kv(k, v):
-    smt = SparseMerkleTree(keysize=1)
-    
+    smt = SparseMerkleTree(key_size=len(k))
+
     # Nothing has been added yet
     assert not smt.exists(k)
-    
+
     # Now that something is added, it should be consistent
     smt.set(k, v)
     assert smt.get(k) == v
     assert smt.root_hash == calc_root(k, v, smt.branch(k))
-    
+
     # If you delete it, it goes away
     smt.delete(k)
     assert not smt.exists(k)
+
+
+@given(
+    data=st.data(),
+    key_size=st.integers(min_value=1, max_value=32),
+)
+def test_branch_updates(data, key_size):
+    num_elements = data.draw(st.integers(min_value=1, max_value=key_size*8//10+1))
+    keys = data.draw(
+        st.lists(
+            elements=st.binary(min_size=key_size, max_size=key_size),
+            min_size=1,
+            max_size=num_elements,
+            unique=True,
+        )
+    )
+    vals = data.draw(
+        st.lists(
+            elements=st.binary(min_size=1, max_size=32),
+            min_size=len(keys),
+            max_size=len(keys),
+        )
+    )
+
+    # Empty tree
+    smt = SparseMerkleTree(key_size=key_size)
+
+    # NOTE: smt._get internal method is used for testing only
+    EMPTY_NODE_HASHES = list(smt._get(keys[0])[1])
+
+    # Objects to track proof data
+    proofs = dict([(k, SparseMerkleProof(k, BLANK_NODE, EMPTY_NODE_HASHES)) for k in keys])
+
+    # Track the big list of all updates
+    proof_updates = []
+    for k, p in proofs.items():
+        # Update the key in the smt a bunch of times
+        for v in vals:
+            proof_updates.append((k, v, smt.set(k, v)))
+
+        # Merge all of the updates into the tracked proof entries
+        for u in proof_updates:
+            p.merge(*u)
+
+        # All merges should be consistent with the latest smt root
+        assert p.root_hash == smt.root_hash
--- a/tests/test_smt.py
+++ b/tests/test_smt.py
@@ -33,32 +33,29 @@ def test_simple_kv(k, v):
     assert not smt.exists(k)
     assert smt.root_hash == empty_root
 
+
 @given(
-    data=st.data(),
-    key_size=st.integers(min_value=1, max_value=32),
+    key_size = st.shared(st.integers(min_value=1, max_value=32), key="key_size"),
+    # Do this so that the size of the keys (in bytes) matches the key_size for the test
+    keys = st.shared(st.integers(), key="key_size").flatmap(lambda key_size: st.lists(
+        elements=st.binary(min_size=key_size, max_size=key_size),
+        min_size=3,
+        max_size=3,
+        unique=True,
+    )),
+    vals = st.lists(
+        elements=st.binary(min_size=1, max_size=32),
+        min_size=3,
+        max_size=3,
+    ),
 )
-def test_branch_updates(data, key_size):
-    num_elements = data.draw(st.integers(min_value=1, max_value=key_size*8//10+1))
-    keys = data.draw(
-        st.lists(
-            elements=st.binary(min_size=key_size, max_size=key_size),
-            min_size=1,
-            max_size=num_elements,
-            unique=True,
-        )
-    )
-    vals = data.draw(
-        st.lists(
-            elements=st.binary(min_size=1, max_size=32),
-            min_size=len(keys),
-            max_size=len(keys),
-        )
-    )
+def test_branch_updates(key_size, keys, vals):
 
     # Empty tree
     smt = SparseMerkleTree(key_size=key_size)
 
     # NOTE: smt._get internal method is used for testing only
+    #       because it doesn't do null checks on the empty default
     EMPTY_NODE_HASHES = list(smt._get(keys[0])[1])
 
     # Objects to track proof data
--- a/trie/smt.py
+++ b/trie/smt.py
@@ -198,6 +198,12 @@ class SparseMerkleTree:
         on-chain datastores.
 
         :param key_size: The size (in # of bytes) of the key. All keys must be this size.
+                         Note that the size should be between 1 and 32 bytes. For performance,
+                         it is not advisible to have a key larger than 32 bytes (and you should
+                         optimize to much less than that) but if the data structure you seek
+                         to use as a key is larger, the suggestion would be to hash that
+                         structure in a serialized format to obtain the key, or add a unique
+                         identifier to the structure.
         :param default: The default value used for the database. Initializes the root.
         """
         # Ensure we can support the given depth
--- a/trie/smt.py
+++ b/trie/smt.py
@@ -1,0 +1,193 @@
+from eth_utils import (
+        keccak,
+        to_int,
+    )
+
+from trie.constants import (
+    BLANK_NODE,
+)
+
+from trie.exceptions import (
+        ValidationError,
+    )
+
+from trie.validation import (
+    validate_is_bytes,
+    validate_length,
+)
+
+
+def calc_root(key, value, branch):
+    path = to_int(key)
+    target_bit = 1
+    # traverse the path in leaf->root order
+    # branch is in root->leaf order (key is in MSB to LSB order)
+    node_hash = keccak(value)
+    for sibling in reversed(branch):
+        if path & target_bit:
+            node_hash = keccak(sibling + node_hash)
+        else:
+            node_hash = keccak(node_hash + sibling)
+        target_bit <<= 1
+
+    return node_hash
+
+
+class SparseMerkleTree:
+
+    def __init__(self, keysize=32):
+        # Ensure we can support the given depth
+        if not 1 <= keysize <= 32:
+            raise ValidationError("Keysize must be number of bytes in range [0, 32]")
+
+        self.keysize = keysize  # keysize in bytes
+        self.depth = keysize * 8  # depth is number of bits in key
+        
+        # Initialize an empty tree with one branch
+        self.db = {}
+        node = BLANK_NODE  # Leaf node
+        for i in range(self.depth):
+            node_hash = keccak(node)
+            self.db[node_hash] = node
+            node = node_hash + node_hash
+
+        # Finally, write the root hash
+        self.root_hash = keccak(node)
+        self.db[self.root_hash] = node
+    
+    @classmethod
+    def from_db(cls, db, root_hash, keysize=32):
+        smt = cls(keysize=keysize)
+
+        # If db is provided, and is not consistent,
+        # there may be a silent error. Can't solve that easily.
+        for k, v in db.items():
+            validate_is_bytes(k)
+            validate_length(k, 32)  # Must be a bytes32 hash
+            validate_is_bytes(v)
+            smt.db[k] = v
+
+        # Set root_hash, so we know where to start
+        validate_is_bytes(root_hash)
+        validate_length(root_hash, 32)  # Must be a bytes32 hash
+
+        smt.root_hash = root_hash
+
+        return smt
+    
+    def get(self, key):
+        value, _ = self._get(key)
+
+        # Ensure that it isn't blank!
+        if value == BLANK_NODE:
+            raise KeyError("Key does not exist")
+        
+        return value
+    
+    def branch(self, key):
+        value, branch = self._get(key)
+
+        # Ensure that it isn't blank!
+        if value == BLANK_NODE:
+            raise KeyError("Key does not exist")
+        
+        return branch
+
+    def _get(self, key):
+        """
+        Returns db value and branch in root->leaf order
+        """
+        validate_is_bytes(key)
+        validate_length(key, self.keysize)
+        branch = []
+
+        target_bit = 1 << (self.depth - 1)
+        path = to_int(key)
+        node_hash = self.root_hash
+        # Append the sibling to the branch
+        # Iterate on the parent
+        for i in range(self.depth):
+            if path & target_bit:
+                branch.append(self.db[node_hash][:32])
+                node_hash = self.db[node_hash][32:]
+            else:
+                branch.append(self.db[node_hash][32:])
+                node_hash = self.db[node_hash][:32]
+            target_bit >>= 1
+
+        # Value is the last hash in the chain
+        # NOTE: Didn't do exception here for testing purposes
+        return self.db[node_hash], branch
+
+    def set(self, key, value):
+        """
+        Returns all updated hashes in root->leaf order
+        """
+        validate_is_bytes(key)
+        validate_length(key, self.keysize)
+        validate_is_bytes(value)
+
+        path = to_int(key)
+        node = value
+        _, branch = self._get(key)
+        proof_update = []  # Keep track of proof updates
+
+        target_bit = 1
+        # branch is in root->leaf order, so flip
+        for sibling in reversed(branch):
+            # Set
+            node_hash = keccak(node)
+            proof_update.append(node_hash)
+            self.db[node_hash] = node
+
+            # Update
+            if (path & target_bit):
+                node = sibling + node_hash
+            else:
+                node = node_hash + sibling
+
+            target_bit <<= 1
+
+        # Finally, update root hash
+        self.root_hash = keccak(node)
+        self.db[self.root_hash] = node
+
+        # updates need to be in root->leaf order, so flip back
+        return list(reversed(proof_update))
+    
+    def exists(self, key):
+        validate_is_bytes(key)
+        validate_length(key, self.keysize)
+
+        try:
+            self.get(key)
+            return True
+        except KeyError:
+            return False
+
+    def delete(self, key):
+        """
+        Equals to setting the value to None
+        Returns all updated hashes in root->leaf order
+        """
+        validate_is_bytes(key)
+        validate_length(key, self.keysize)
+
+        return self.set(key, BLANK_NODE)
+    
+    #
+    # Dictionary API
+    #
+    
+    def __getitem__(self, key):
+        return self.get(key)
+    
+    def __setitem__(self, key, value):
+        self.set(key, value)
+    
+    def __delitem__(self, key):
+        self.delete(key)
+    
+    def __contains__(self, key):
+        return self.exists(key)
+
--- a/trie/smt.py
+++ b/trie/smt.py
@@ -189,7 +189,9 @@ class SparseMerkleProof:
 
 class SparseMerkleTree:
 
-    def __init__(self, key_size: int=32, default: bytes=BLANK_NODE):
+    def __init__(self,
+                 key_size: int = 32,
+                 default: bytes = BLANK_NODE):
         """
         Maintain a a binary trie with a particular depth (defined by key size)
         All values are stored at that depth, and the tree has a default value that it is
@@ -232,8 +234,8 @@ class SparseMerkleTree:
             cls,
             db: Dict[bytes, bytes],
             root_hash: Hash32,
-            key_size: int=32,
-            default: bytes=BLANK_NODE):
+            key_size: int = 32,
+            default: bytes = BLANK_NODE):
         smt = cls(key_size=key_size, default=default)
 
         # If db is provided, and is not consistent,
--- a/trie/smt.py
+++ b/trie/smt.py
@@ -4,8 +4,6 @@ from typing import (
     Tuple,
 )
 
-import copy
-
 from eth_utils import (
     keccak,
     to_int,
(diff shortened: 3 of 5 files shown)


trie/smt.py
```diff
@@ -0,0 +1,177 @@
...
+    validate_is_bytes,
+    validate_length,
+)
+
+
+class SparseMerkleTree:
+
+    def __init__(self, keysize=32):
+        # Ensure we can support the given height
+        if 1 > keysize or keysize > 32:
```
pipermerriam: Maybe more legible:

```
if not 0 <= keysize <= 32:
   ...
```

trie/smt.py
```diff
@@ -0,0 +1,177 @@
...
+            raise ValidationError("Keysize must be number of bytes in range [0, 32]")
+
+        self.keysize = keysize  # keysize in bytes
+        self.height = keysize * 8  # height is number of bits in key
```
pipermerriam: Is `height` a term from the spec.  I think `depth` might be another term we use for this measurement and for me at least it's a more natural name.  I don't feel strongly, but I'd default to pushing for `depth` absent an objective reason to use something else.

trie/smt.py
```diff
@@ -0,0 +1,177 @@
...
+        
+        # Initialize an empty tree with one branch
+        self.db = {}
```
pipermerriam, line 221: I'd suggest following the same pattern as `HexaryTrie` and having the `db` be something that is passed in (so we can use non-memory based databases).

pipermerriam (commented):
Just some comments, didn't go through any of the core methods yet.

trie/smt.py
fubuloubu: `depth` :+1:

trie/smt.py
fubuloubu: I built a class method for that purpose, since you have to initialize the database to a particular depth, with the root hash at the top. I could have that in the constructor, but I wanted to keep it separate to make it clearer that there is the potential for inconsistency if it accepts an outside db.

fubuloubu:
Ready for review

fubuloubu:
Ready for Review

EDIT: Added a class to manage Merkle Proofs in a more useful way. The intention would be to subclass this manager for different use cases (i.e. non-bytes types for keys and values)

trie/smt.py
```diff
@@ -0,0 +1,266 @@
...
+
+        self._default = default
+
+        # Initialize an empty tree with one branch
+        self.db = {}
+        node = self._default  # Default leaf node
+        for i in range(self.depth):
+            node_hash = keccak(node)
+            self.db[node_hash] = node
+            node = node_hash + node_hash
```
pipermerriam, line 226: Experimented with how to do this without overriding the local `node` variable on each iteration.

```
    def do_iteration(value):
        node_hash = keccak(value)
        db[node_hash] = value
        return node_hash + node_hash

    root_node = pipe(
        node,
        *itertools.repeat(do_iteration, depth),
    )
    ...
```
    
Readability is reasonably poor so not a clear winner.

trie/smt.py
```diff
@@ -0,0 +1,266 @@
...
+                node = node_hash + sibling
+
+            target_bit <<= 1
+
+        # Finally, update root hash
+        self.root_hash = keccak(node)
+        self.db[self.root_hash] = node
+
+        # updates need to be in root->leaf order, so flip back
+        return list(reversed(proof_update))
```
pipermerriam: We've tended to use `tuple` unless we have an explicit reason to need mutability.

trie/smt.py
```diff
@@ -0,0 +1,266 @@
...
+        validate_is_bytes(key)
+        validate_length(key, self.keysize)
+        branch = []
+
+        target_bit = 1 << (self.depth - 1)
+        path = to_int(key)
+        node_hash = self.root_hash
+        # Append the sibling to the branch
+        # Iterate on the parent
+        for i in range(self.depth):
```
pipermerriam: We have tended to use `_` for loop variables that we don't actually make use of.

trie/smt.py
```diff
@@ -0,0 +1,266 @@
...
+
+
+class SparseMerkleProof:
+
+    def __init__(self, key, value, branch):
+        validate_is_bytes(key)
+        self._key = key
+        self._keysize = len(key)
+        self._value = value
+        validate_length(branch, self._keysize*8)
```
pipermerriam: Seems like it might be better to group both of the validation statements together at the beginning of the function body.

trie/smt.py
```diff
@@ -0,0 +1,266 @@
...
+        self._branch = deepcopy(branch)
+        self._branchsize = len(branch)
```
pipermerriam: naming convention would be to have this as `_branch_size`

trie/smt.py
```diff
@@ -0,0 +1,266 @@
...
+
+    return node_hash
+
+
+class SparseMerkleProof:
+
+    def __init__(self, key, value, branch):
+        validate_is_bytes(key)
+        self._key = key
+        self._keysize = len(key)
```
pipermerriam: naming convention would have this be `_key_size`

pipermerriam (commented):
I'm still sort of struggling to understand how this would be used.  Maybe I missed it somewhere.  Are there code examples that might be illustrative here?

trie/smt.py
fubuloubu: Will make this update

trie/smt.py
fubuloubu: Will do

trie/smt.py
fubuloubu: Does this mean I should pre-compute the number of bits the key contains?

trie/smt.py
fubuloubu: Will do

trie/smt.py
fubuloubu: Will do

fubuloubu:
@pipermerriam I assume you mean the `SparseMerkleProof` class (SMT should have pretty clear utility)

Here is an example of me using it in practice:
https://github.com/GunClear/PlasmaRifle/blob/9fc9c4ad6510b7f1b5a9fdcb2698c973d367002b/demo/get-branch.py#L56-L102

This class is listening to an address on the public network in order to track updates in an on-chain managed merkle tree. This contract doesn't store the full database in state, but communicates node updates to any party that iterates through the logs to synchronize their locally maintained branch to the contract. This saves some gas because the node updates are stored in logs vs. state (ephermality of logs is acceptable). The end result is that the user has a merkle proof they can submit to prove membership in that list and do things on- or off-chain. My use case requires this to generate a ZKP of membership in this on-chain list without leaking identity in order to authorize transactions.

Another use case would be for Plasma Cash designs that have state-enabled checkpointing. When a user wants to exit a token in this design, they would need to upload a proof of membership of that token in the state tree tracked by the on-chain list of state root hashes such a design would manage. I could picture something similar being of utility for protocol layer clients. Note that I was mistaken that this functionality would be useful for general Plasma Cash designs as they only track the differences between blocks (e.g. the transaction root instead of the state root)

trie/smt.py
fubuloubu: So... should I do anything to make this better?

fubuloubu:
Failure was because this project is using an old version of hypothesis (3.7.0 was Mar 2017, 3.82.1 is latest)

tests/test_smt.py
```diff
@@ -0,0 +1,79 @@
...
+    # If you delete it, it goes away
+    smt.delete(k)
+    assert not smt.exists(k)
+
+
+@given(
+    data=st.data(),
+    key_size=st.integers(min_value=1, max_value=32),
+)
+@settings(deadline=None)
```
fubuloubu: This line does not work in hypothesis 3.7.0

trie/smt.py
```diff
@@ -0,0 +1,355 @@
...
+
+    .. doctest::
+
+        >>> # smt is located on another process or machine
+        >>> smt = SparseMerkleTree(key_size=1)
+        >>> our_key = b'\x03'
+        >>> our_value = b'\x01'
+        >>> smt.set(our_key, our_value)
+        >>> # We need to track proof data for *some* reason
+        >>> our_proof = SparseMerkleProof(our_key, our_value, smt.branch(our_key))
```
fubuloubu, line 90: Added a doctest to show off how this would be used

trie/smt.py
pipermerriam: No, literally just moving this statement up a few lines prior to setting the various values on `self`

tests/test_constants.py
```diff
@@ -0,0 +1,39 @@
...
+    DEPTH = 256  # Default depth is 32 bytes
+
+    # Start at the bottom
+    EMPTY_LEAF_NODE_HASH = BLANK_HASH
+    EMPTY_NODE_HASHES = [EMPTY_LEAF_NODE_HASH]
+
+    # More hashes the lower you go down the tree (to the root)
+    # NOTE: Did this with different code as a sanity check
+    for _ in range(DEPTH - 1):
+        EMPTY_NODE_HASHES.insert(0, keccak(EMPTY_NODE_HASHES[0] + EMPTY_NODE_HASHES[0]))
```
pipermerriam: FYI, `list.insert(0, thing)` is one of the slowest standard library operations around.  My recommendation would be to either:

- build the list in reverse order and then reverse it once you are done
- use the `toolz.accumulate` function which does exactly what you need here
- use a `collections.deque` which has fast poping and insertion at both ends.

tests/test_smt.py
```diff
@@ -0,0 +1,79 @@
...
+    # If you delete it, it goes away
+    smt.delete(k)
+    assert not smt.exists(k)
+
+
+@given(
+    data=st.data(),
+    key_size=st.integers(min_value=1, max_value=32),
+)
+#@settings(deadline=None)  # This option is not available in hypothesis-3.7.0
```
pipermerriam: This like should just be removed.

trie/smt.py
pipermerriam: > So... should I do anything to make this better?

I don't think so.  Most of the time when I see mutation I like to experiment with what it takes to remove it.  Often it isn't worth it (like this case) but I still like dumping the answer in case it's educational to anyone for how to do certain things using an functional approach.

pipermerriam (approved):

tests/test_smt.py
fubuloubu: It found a pretty decent bug I wouldn't have found otherwise, but can remove if upgrading isn't worth hassle.

tests/test_constants.py
fubuloubu: I will do this update, and try to squash some of it down.

fubuloubu: Used `collections.deque`, but didn't really make a difference

fubuloubu:
Merged all commits into 3. Everything looks good to go!

tests/test_smt.py
```diff
@@ -0,0 +1,77 @@
...
+    assert not smt.exists(k)
+
+    # Now that something is added, it should be consistent
+    smt.set(k, v)
+    assert smt.get(k) == v
+    assert smt.root_hash == calc_root(k, v, smt.branch(k))
+
+    # If you delete it, it goes away
+    smt.delete(k)
+    assert not smt.exists(k)
```
carver, line 33: Maybe also confirm that the root hash is the same as it started?

tests/test_constants.py
```diff
@@ -0,0 +1,41 @@
...
+
+    # Start at the bottom
+    EMPTY_LEAF_NODE_HASH = BLANK_HASH
+    EMPTY_NODE_HASHES = collections.deque([EMPTY_LEAF_NODE_HASH])
+
+    # More hashes the lower you go down the tree (to the root)
+    # NOTE: Did this with different code as a sanity check
+    for _ in range(DEPTH - 1):
+        EMPTY_NODE_HASHES.appendleft(keccak(EMPTY_NODE_HASHES[0] + EMPTY_NODE_HASHES[0]))
+    EMPTY_ROOT_HASH = keccak(EMPTY_NODE_HASHES[0] + EMPTY_NODE_HASHES[0])
```
carver, line 34: I think you only need to calculate this 32byte empty root hash once, and then save it as a constant to check against in the test (probably useful to have around anyhow). Putting the code in a utility for later verification/re-generation seems sufficient.

tests/test_smt.py
```diff
@@ -0,0 +1,77 @@
...
+    num_elements = data.draw(st.integers(min_value=1, max_value=key_size*8//10+1))
+    keys = data.draw(
+        st.lists(
+            elements=st.binary(min_size=key_size, max_size=key_size),
+            min_size=1,
+            max_size=num_elements,
+            unique=True,
+        )
+    )
+    vals = data.draw(
```
carver: I haven't used data.draw before, but it seems like it makes it more difficult to reproduce the scenario if the parameters are generated at runtime instead of before the test. Doing it this other way also makes me wonder: is there a reason the standard way wouldn't work? (in this case: adding `num_elements`, `keys`, and `vals` as more parameters to `@given`)

tests/test_smt.py
```diff
@@ -0,0 +1,77 @@
...
+    # NOTE: smt._get internal method is used for testing only
+    EMPTY_NODE_HASHES = list(smt._get(keys[0])[1])
+
+    # Objects to track proof data
+    proofs = dict([(k, SparseMerkleProof(k, BLANK_NODE, EMPTY_NODE_HASHES)) for k in keys])
+
+    # Track the big list of all updates
+    proof_updates = []
+    for k, p in proofs.items():
+        # Update the key in the smt a bunch of times
```
carver, line 67: Why a bunch of times here? Why specifically are we doing this the number if times that there are keys? Is that a coincidence? I'm not clear why there are multiple resets of the values on the same key beyond testing one or two.

tests/test_smt.py
```diff
@@ -0,0 +1,77 @@
...
+        for v in vals:
+            proof_updates.append((k, v, smt.set(k, v)))
+
+        # Merge all of the updates into the tracked proof entries
+        for u in proof_updates:
+            p.merge(*u)
```
carver: Generating the proof requires re-applying the relevant keys and values and the result of `smt.set()`?

carver:
Cool, taking a look!

trie/smt.py
```diff
@@ -0,0 +1,355 @@
...
+    ValidationError,
+)
+
+from trie.validation import (
+    validate_is_bytes,
+    validate_length,
+)
+
+
+def calc_root(key: bytes, value: bytes, branch: Sequence[Hash32]) -> Hash32:
```
carver, line 30: Maybe `siblings` instead of `branch`?

trie/smt.py
```diff
@@ -0,0 +1,355 @@
...
+    Obtain the merkle root of a given key/value/branch set.
+    Can be used to validate a merkle proof or compute it's value from data.
+
+    :param key: the keypath to decide the ordering of the siblings in the branch
+    :param value: the value (or leaf) that starts the merkle proof computation
+    :param branch: the sequence of siblings used to recursively perform the computation
+
+    :return: the root hash of the merkle proof computation
+
+    .. doctest::
```
carver, line 41: Neat, is the repo set up to run doctests?

trie/smt.py
```diff
@@ -0,0 +1,355 @@
...
+                raise ValidationError("Updated node list is not deep enough")
+
+            # Update sibling in the branch where our key differs from the update
+            self._branch[branch_point] = node_updates[branch_point]
+            # No need to update value
+
+
+class SparseMerkleTree:
+
+    def __init__(self, key_size: int=32, default: bytes=BLANK_NODE):
```
carver: Worth mentioning in a docstring that `key_size` is in bytes.

trie/smt.py
```diff
@@ -0,0 +1,355 @@
...
+
+    return node_hash
+
+
+class SparseMerkleProof:
+    """
+    Track the current value and merkle proof branch for a given key.
+    This will enable the tracked proof data to stay up to date with changes
+    that may be streamed to the end user over external protocols without having
+    to interactively query the full SMT to obtain the most up-to-date branch.
```
carver, line 74: I can see the value in not *having* to query the full SMT, but it seems like an option that should be supported. (I think it's fine to add an issue as a TODO, so this PR doesn't drag on).

It's worth noting that if this were used to generate proofs locally that it ~doubles the amount of hashing done, which is usually a significant computation in aggregate.

trie/smt.py
```diff
@@ -0,0 +1,355 @@
...
+        >>> smt.set(our_key, our_value)
+        >>> # We need to track proof data for *some* reason
+        >>> our_proof = SparseMerkleProof(our_key, our_value, smt.branch(our_key))
+        >>> their_key = b'\x05'
+        >>> their_new_value = b'\x01'
+        >>> their_node_updates = smt.set(their_key, their_new_value)
+        >>> # tree updates can be communicated over any channel to proof obj
+        >>> our_proof.merge(their_key, their_new_value, their_node_updates)
+        >>> # Note our branch data was never queried from smt to our proof obj
+        >>> our_proof.branch == smt.branch(our_key)
```
carver, line 97: I think `siblings` could be a good name for both this attribute and method, too.

trie/smt.py
```diff
@@ -0,0 +1,355 @@
...
+        >>> our_key = b'\x03'
+        >>> our_value = b'\x01'
+        >>> smt.set(our_key, our_value)
+        >>> # We need to track proof data for *some* reason
+        >>> our_proof = SparseMerkleProof(our_key, our_value, smt.branch(our_key))
+        >>> their_key = b'\x05'
+        >>> their_new_value = b'\x01'
+        >>> their_node_updates = smt.set(their_key, their_new_value)
+        >>> # tree updates can be communicated over any channel to proof obj
+        >>> our_proof.merge(their_key, their_new_value, their_node_updates)
```
carver: `merge` to me usually means combining two things of equal type. Maybe `update()`? (like `dict`'s)

trie/smt.py
```diff
@@ -0,0 +1,355 @@
...
+
+    def __init__(self, key: bytes, value: bytes, branch: Sequence[Hash32]):
+        validate_is_bytes(key)
+        validate_is_bytes(value)
+        validate_length(branch, len(key)*8)
+
+        self._key = key
+        self._key_size = len(key)
+        self._value = value
+        self._branch = list(copy.deepcopy(branch))  # Avoid issues with mutable lists
```
carver: `Hash32` isn't mutable, so isn't `self._branch = list(siblings)` sufficient?

trie/smt.py
```diff
@@ -0,0 +1,355 @@
...
+    def value(self) -> bytes:
+        return self._value
+
+    @property
+    def branch(self) -> Tuple[Hash32]:
+        return tuple(self._branch)
+
+    @property
+    def root_hash(self) -> Hash32:
+        return calc_root(self.key, self.value, self.branch)
```
carver, line 142: Probably want this in a cache, but I'm happy if that just goes in a follow-up issue.

trie/smt.py
```diff
@@ -0,0 +1,355 @@
...
+        else:
+            # Find the first mismatched bit between keypaths. This is
+            # where the branch point occurs, and we should update the
+            # sibling node in the source branch at the branch point.
+            # NOTE: Keys are in MSB->LSB (root->leaf) order.
+            #       Node lists are in root->leaf order.
+            #       Be sure to convert between them effectively.
+            for bit in reversed(range(self._branch_size)):
+                if path_diff & (1 << bit) > 0:
+                    branch_point = (self._branch_size-1)-bit
```
carver: Yikes, how is `flake8` not complaining? It seems like it's running, but it should have required spaces around `-`.

trie/smt.py
```diff
@@ -0,0 +1,355 @@
...
+                    break
+
+            # NOTE: node_updates only has to be as long as necessary
+            #       to obtain the update. This allows an optimization
+            #       of pruning updates to the maximum possible depth
+            #       that would be required to merge, which may be many
+            #       orders of magnitude smaller than the tree depth.
```
carver: "many orders of magnitude" sounds like a stretch. 256 values is only ~2 orders of magnitude larger than 1. :) "significantly" seems a sufficient word here.

trie/smt.py
```diff
@@ -0,0 +1,355 @@
...
+            self._branch[branch_point] = node_updates[branch_point]
+            # No need to update value
+
+
+class SparseMerkleTree:
+
+    def __init__(self, key_size: int=32, default: bytes=BLANK_NODE):
+        # Ensure we can support the given depth
+        if not 1 <= key_size <= 32:
+            raise ValidationError("Keysize must be number of bytes in range [0, 32]")
```
carver: Why are we limited to 32? Not really important, but I don't see anything that prevents longer keys, yet.

trie/smt.py
```diff
@@ -0,0 +1,355 @@
...
+
+        self._key_size = key_size  # key size in bytes
+        self.depth = key_size * 8  # depth is number of bits in key
+
+        self._default = default
+
+        # Initialize an empty tree with one branch
```
carver, line 220: Ok, this time it can be a branch. :)

trie/smt.py
```diff
@@ -0,0 +1,355 @@
...
+            cls,
+            db: Dict[bytes, bytes],
+            root_hash: Hash32,
+            key_size: int=32,
+            default: bytes=BLANK_NODE):
+        smt = cls(key_size=key_size, default=default)
+
+        # If db is provided, and is not consistent,
+        # there may be a silent error. Can't solve that easily.
+        for key, value in db.items():
```
carver: I don't think we can assume that the database is iterable. (or we often wouldn't want to, even if it is possible)

I suspect this can be modified to keep a reference to the database, and look up all the values on demand (like the other tries do).

trie/smt.py
```diff
@@ -0,0 +1,355 @@
...
+        return smt
+
+    def get(self, key: bytes) -> bytes:
+        value, _ = self._get(key)
+
+        # Ensure that it isn't blank!
+        if value == BLANK_NODE:
+            raise KeyError("Key does not exist")
+
+        return value
```
carver, line 260: Just a nicety if you like. We've found some success in doing `if/elif` tries even when they aren't *required* by control flow, like:

```py
        if value == BLANK_NODE:
            raise KeyError("Key does not exist")
        else:
            return value
```

trie/smt.py
```diff
@@ -0,0 +1,355 @@
...
+        node_hash = self.root_hash
+        # Append the sibling to the branch
+        # Iterate on the parent
+        for _ in range(self.depth):
+            if path & target_bit:
+                branch.append(self.db[node_hash][:32])
+                node_hash = self.db[node_hash][32:]
+            else:
+                branch.append(self.db[node_hash][32:])
+                node_hash = self.db[node_hash][:32]
```
carver: Maybe you would like:

```py
        for _ in range(self.depth):
            node = self.db[node_hash]
            left, right = node[:32], node[32:]

            if path & target_bit:
                node_hash = right
                siblings.append(left)
            else:
                node_hash = left
                siblings.append(right)

**Review by carver — CHANGES_REQUESTED** [review]
Ok, *almost* got through it before I had to take off. No huge problems, but I'll add an X since Piper already gave it a :heavy_check_mark: , as a reminder that there are still a few things to go over.

I really think you're going to want to do the db like the other trie's do, to make it easy to have a disk-backed db.

Also, you'll probably want something like `squash_changes` so that you don't flood the db with a bunch of intermediate state that you don't care about in the middle of a transaction:
https://github.com/ethereum/py-trie/blob/bce9e027b0ef4f411c2eabc0600e6b94679075d8/trie/hexary.py#L444-L450

File: trie/smt.py (hunk 24 above)
Comment (fubuloubu) [reply trie/smt.py]: I mean, they are sort of equal type, but also not really.

`update` is good enough haha

File: trie/smt.py (hunk 23 above)
Comment (fubuloubu) [reply trie/smt.py:97]: Branch has a particular technical meaning though. "Siblings" isn't entirely accurate because they are at different levels in the tree.

Could brainstorm better words.

File: trie/smt.py (hunk 25 above)
Comment (fubuloubu) [reply trie/smt.py]: Fair. Reduces dependencies

File: trie/smt.py (hunk 26 above)
Comment (fubuloubu) [reply trie/smt.py:142]: You would probably check this only once before using it, which would update the cached value.

Not sure if a caching strategy would make sense, but open to seeing how the use cases play out.

File: trie/smt.py (hunk 28 above)
Comment (fubuloubu) [reply trie/smt.py]: Lol fair enough :)

File: trie/smt.py (hunk 29 above)
Comment (fubuloubu) [reply trie/smt.py]: That is true, but anything above 256 bit depth is impractical IMO. If it was suggested for a use case, I would counter by saying "hash the key"

File: trie/smt.py (hunk 31 above)
Comment (fubuloubu) [reply trie/smt.py]: Can you point to an example?

Also, what do you think about this being a separate class method?

File: trie/smt.py (hunk 32 above)
Comment (fubuloubu) [reply trie/smt.py:260]: Does that lead to an optimization or better code analysis?

File: trie/smt.py (hunk 33 above)
Comment (fubuloubu) [reply trie/smt.py]: Yeah, that's clearer.

File: tests/test_smt.py (hunk 16 above)
Comment (fubuloubu) [reply tests/test_smt.py]: I couldn't figure out how to make it work where the size of the keys in `keys` depended on the choice of `key_size` pulled from that strategy. I also needed the size of `vals` to match the size of `keys`, which also must be less than the number of possible options for the given `key_size`, although I suppose `unique=True` also guarantees this...

Open to suggestions on how to do that.

File: tests/test_smt.py (hunk 17 above)
Comment (fubuloubu) [reply tests/test_smt.py:67]: Yeah, I could see that argument. I wanted to show that even with sufficient amount of updates there is no loss of correctness. But this could occur with even 3 selections. Combined with the above issue with strategies, I could probably reduce this so that it's doing less unnecessary verification by having a hard coded number of updates.

File: tests/test_smt.py (hunk 18 above)
Comment (fubuloubu) [reply tests/test_smt.py]: `smt.set()` returns what the node updates are for a given change in value in the tree. These updates can be synchronized with a separate proof object (perhaps on a different computer), and that proof object will find the relevant sibling to keep itself up to date so that the proof object can track the same smt root separately without the extra data overhead of managing the whole tree (and the stream of all key-value updates). The storage size of managing the tree can grow non-linearly, whereas the proof object size will remain constant. This is at the expense of increased networking overhead, but that can be optimized by only communicating these updates to the practical depth of the tree (beyond which all keys map to single-branch sub-trees).

My intended use case is mobile, so this tradeoff of storage size is desirable. If my use case had 1m users, the tree would grow to be 2+ gigs in size (without any optimizations).

File: trie/smt.py (hunk 19 above)
Comment (fubuloubu) [reply trie/smt.py:30]: branch is a technical term, even the binary trie uses it: https://github.com/ethereum/py-trie/blob/master/trie/binary.py#L254

File: trie/smt.py (hunk 20 above)
Comment (fubuloubu) [reply trie/smt.py:41]: Nope lol.

I could turn this into an actual test case

File: tests/test_smt.py (hunk 13 above)
Comment (pipermerriam) [reply tests/test_smt.py]: My comment about removing it is that it commenting it out is equivalent to removing it but with the negative downside of leaving it in the codebase.  If we're leaving it in it should get a comment explaining why.  If we are no longer using hypothesis versions that support this option I don't see why we should leave this in...

File: tests/test_smt.py (hunk 16 above)
Comment (pipermerriam) [reply tests/test_smt.py]: https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.shared

I think this is what you want.

```python
@given(
    key_size=st.shared(st.integers(min_value=1, max_value=32), key="key_size"),
    kv_pairs=st.shared(st.integers(min_value=1, max_value=32), key="key_size").flatmap(
       lambda key_size: st.tuples(st.binary(min_size=key_size, max_size=key_size), st.binary(min_size=1, max_size=1))
    ),
)
def test_thing(key_size, kv_pairs):
    ...
```
)

pipermerriam: ^ the above is psuedo-code but should get you going in the right direction.

trie/smt.py
carver: It's all about readability: the visual confirmation that this is a branching situation and the lines below it only happen in the alternate case.

trie/smt.py
carver: > Can you point to an example?

Say, all of the state in Ethereum mainnet. Even though plyvel technically supports key iteration it would be prohibitively expensive ~~computationally~~ in I/O.

> Also, what do you think about this being a separate class method?

I think if this becomes like the other tries, with a reference to the underlying db, then most of this method can disappear, and the rest moves to init.

trie/smt.py
carver: Ok, maybe a comment would help, saying something like "the 32-byte limit is to purely encourage good performance characteristics"

tests/test_smt.py
carver: Sorry, this was a comment from before I read the whole thing. Meant to delete it, but it snuck in on my way out the door. Totally on board. :+1:

trie/smt.py
carver: That's a branch *node*, to distinguish it from an extension or leaf node. (Perhaps I would have suggested `branching_node` instead, but not worth rehashing at this point)

A branch in a tree is the list of ancestors from the leaf to the root. The argument in this function is the list of the siblings of the ancestors.

One confirmation that it's a good name is that a loop below would change from:

```py
for sibling in branch:
```
to:

```py
for sibling in siblings:
```

trie/smt.py
carver: > Branch has a particular technical meaning though.

Okay, maybe I'll take back my definition of branch from above, I couldn't find any supporting sources. (Although I am not sure which technical meaning you're referring to either)

> "Siblings" isn't entirely accurate because they are at different levels in the tree.

Sure, in my head it's short for "siblings for the key nodes", or more pedantically "siblings for the nodes corresponding to each bit in the key". I still don't think it's a terrible name, but I'm happy to keep brainstorming.

Other options
- `uncles`
- `ancestor_uncles`
- `ancestor_siblings`
- `pruned_siblings`
- `proof_siblings`
- `proof_nodes`
- `key_siblings`

One reason why I still include `siblings`: the docstring and for loop both use sibling to refer to an individual node in the list.

trie/smt.py
carver: Yup, totally happy to just punt :football:

tests/test_smt.py
fubuloubu: Ahhhh, that's what the hell `st.shared` does!

Was trying to figure this out on the plane lol

trie/smt.py
fubuloubu: Well, unlike the Ethereum tries, there are two characteristics that I think warrant the separate approach.
1. SMTs can be used in a more ad-hoc fashion.  For example in Plasma, each transaction in a block is bundled into an SMT and summarized via the root, which is uploaded to Ethereum. Probably not too different from the txn root in Ethereum, but wortth noting.
2. You may want to specify a different default value for an empty database to the trie, so that it can be used for different characteristics. For example, storage in Ethereum assumes a default value for the key, which is `bytes32(0)` instead of `b''`. I was trying to allow for this behavior, and disambiguate it from db storage use cases through the separate class method.

trie/smt.py
```diff
@@ -0,0 +1,355 @@
...
+        self.depth = key_size * 8  # depth is number of bits in key
+
+        self._default = default
+
+        # Initialize an empty tree with one branch
+        self.db = {}
+        node = self._default  # Default leaf node
+        for _ in range(self.depth):
+            node_hash = keccak(node)
+            self.db[node_hash] = node
```
carver, line 225: One optimization (probably to go in an issue for later): it looks like these node hashes could be stored in a blank hash ancestor lookup so the `get()` can shortcut and return a blank result as soon as it hits one of the blank hash ancestors. (making it `O(log(h))` instead of `O(h)` operations to find a blank key)

trie/smt.py
carver: On 1) nothing stops you from passing in a {} as the "underlying database", that's a totally reasonable use case that py-evm does sometimes, too.

On 2) if we want to be careful about it, `__init__` could always try to add the default values to the DB (maybe checking if they exist first). Other than that, it seems like we could still follow the same pattern for init in the other tries.

carver:
So `squash_changes` reminded me that there are often a lot of trie changes, even during a single transaction. Which brings up an interesting point: how well does this streaming optimization work, and in what parameters? (and what is the key metric to optimize for?)

For example, if we are trying to optimize for network traffic, it might be better for the SMT to re-broadcast the key proof at every checkpoint.

In this construction, each of those key changes and proofs would have to be calculated and broadcast. Usually, there are only important checkpoints at which the proof is relevant. Say at the end of a block. To play that through a little bit:

Let's say there are 1000 account updates during the block. In the best case, `len(key_siblings) == 1`, meaning all changes are on the other half of the trie. Then there are 1k updates on the order of 32 bytes (ignoring key and value), for a total of 32kb traffic per block. If the SMT broadcasts the new proof for a given key at every block, then it sends (worst case) 32bytes * 256 levels or about 1/4 of the network load.

In this example, you could be even smarter and send only the key siblings that changed, for a total of 32bytes * 1 level, for 1000x improvement. In fact, I think this optimization makes the solution always cost less than or equal to the network usage of the stream of key changes.

---

Unrelated: also note that db size is growing without bound. You might consider using pruning like `HexaryTrie(is_pruning=True)` works and have the `SparseMerkleProof` prune the data along the way, so that your memory/disk doesn't grow unbounded.

fubuloubu:
The Plasma use case and my intended use I think are sufficiently different that we should probably talk offline about optimizations and other attributes of this feature.

The TL;DR is that it's less about efficient storage of the underlying data and more about having deterministic and cheap algorithms for verifying inclusion of data in a data set in something like a smart contract. I could be missing something though, so let's chat about it

trie/smt.py
davesque: @carver Maybe too many cooks here, but isn't it also common for guarding (i.e. stack popping) if statements to lack else clauses?  I know that I often write code in this pattern and sort of prefer it since it's not hard to tell at a glance that the body of the if statement would prevent any further execution in that subroutine.

tests/test_constants.py
fubuloubu: So, the root would change with different choices of the default. I'm not sure a utility would be a good suggestion, but I'm open to that. I wanted to calculate it here in a different way so that that test does not depend on the implementation. To put the branch explicitly would be almost 260 hardcoded lines, so I though this was a good compromise to doing that.

Also, we have the `smt.calc_root(k, v, b)` utility that can obtain a root hash.

trie/smt.py
fubuloubu: Re: 1) {} would not be a valid database, because it does not have a pathway to the default value.
Re: 2) because of 1), not sure if mixing this logic is the best choice because that would have an eiher/or constructor layout

trie/smt.py
fubuloubu: Not sure what you mean

tests/test_constants.py
fubuloubu: So, the root would change with different choices of the default. I'm not sure a utility would be a good suggestion, but I'm open to that. I wanted to calculate it here in a different way so that that test does not depend on the implementation. To put the branch explicitly would be almost 260 hardcoded lines, so I though this was a good compromise to doing that.

Also, we have the `smt.calc_root(k, v, b)` utility that can obtain a root hash.

tests/test_smt.py
fubuloubu: Figured this out, thanks @pipermerriam!

tests/test_smt.py
fubuloubu: Partially made this update by reducing the size of the list to a hard 3. Keeping the loop, because that means the update gets checked before, during, and after an update from another key. Tests are a little faster now :)

trie/smt.py
fubuloubu: `proof` might actually be most accurate. it also shows the use case a little clearer.

trie/smt.py
carver: Yeah, my personal preference is to only do those at the very beginning of a method, rather than after executing some logic. But this is just a light preference, definitely not a required change. `#WONTFIX` is fine

carver:
> it's less about efficient storage of the underlying data and more about having deterministic and cheap algorithms for verifying inclusion of data in a data set in something like a smart contract.

Most of my comment was about efficient network usage, not data storage. When you say "cheap algorithms" which thing are you optimizing for?

trie/smt.py
fubuloubu: So, these examples aren't too substantially different from the test cases to warrant creating more. I could remove it from the docstring to prevent an out-of-date example from weaseling it's way in.

fubuloubu:
> When you say "cheap algorithms" which thing are you optimizing for?

Verification cost in gas

trie/smt.py
carver: 1) huh? `self.db = {}` already in [the PR](https://github.com/ethereum/py-trie/pull/77/files/4f2e5f5e66f553b96bda5b76281a12309b2f16b3#diff-730d445fc351b9302eb71dc655598097R205)
2) I'm also cool with having a separate `smt = SMT.init_db(db, default_value=...)`, where `SMT(db, default_value=...)` would assume the database is already initialized.

fubuloubu: 1. Immediately followed by creating the branch for the default value in the db through the iteration lol
2. Cool, thanks. I think this is clearer that there are two use cases.

trie/smt.py
fubuloubu: I think we can leave this out for now, but definitely something to think about.

fubuloubu:
Also, would really like to have a chat about the networking efficiency/data storage efficiency tradeoff. It makes sense in my head, but perhaps you can help me evaluate a better solution once I communicate the use case better.

carver:
Sure, one thing that would help me is if you could explain why broadcasting proofs at intervals would be sub-optimal for your use case. (or if I am totally misunderstanding the context)

fubuloubu:
Moving to gitter

trie/smt.py
fubuloubu: Solved the original concern by just directly assigning the db in the classmethod.

trie/smt.py
fubuloubu: changed all mentions of `sibling` to `sibling_node`. `branch` is still used to refer to an heirarchically-ordered set of a sibling nodes

trie/smt.py
fubuloubu: As mentioned above, changed all mentions of sibling to sibling_node. branch is still used to refer to an heirarchically-ordered set of a sibling nodes

trie/smt.py
fubuloubu: Not sure what this comment was about, but noted that this could be handled in another PR if further issues are found with the approach.

fubuloubu:
@carver I believe I resolved all of your outstanding comments that you noted should be solved in this version of the PR. Please let me know on here or gitter if you have further questions or comments.

carver:
Okay, yes, I think it is mergeable. I'm still not totally satisfied with the name `branch` and some of the other notes. Since you'd rather stop talking about it, I may decide to change it later without consulting you. :stuck_out_tongue_closed_eyes: 

Of course, that's not even vaguely near the top of my priority heap right now.

When the release notes come out, I'll also note this class as experimental, indicating that the API may change at minor versions.

carver (approved):

trie/smt.py
```diff
@@ -0,0 +1,369 @@
...
+            cls,
+            db: Dict[bytes, bytes],
+            root_hash: Hash32,
+            key_size: int = 32,
+            default: bytes = BLANK_NODE):
+        smt = cls(key_size=key_size, default=default)
+
+        # If db is provided, and is not consistent,
+        # there may be a silent error. Can't solve that easily.
+        smt.db = db
```
carver, line 243: Reminder that we probably want the ability to initialize a passed-in database with empty node hashes (eg~ LevelDB).

PR #80, opened 2019-01-07
Add `get_proof`

zsaladin:
### What was wrong?
- There was not `get_proof` even if `get_from_proof` exists.
- `get_from_proof` had a problem that a trie could not add a node less than 32 bytes as a proof.

### How was it fixed?
- I made `get_proof` method.
- Add proof nodes to db by force in `get_from_proof`

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/tests/test_hexary_trie.py
+++ b/tests/test_hexary_trie.py
@@ -133,13 +133,13 @@ def test_trie_using_fixtures(name, updates, expected, deleted, final_root):
     for valid_proof_key in expected:
         valid_proof = trie.get_proof(valid_proof_key)
         assert len(valid_proof) > 0
-        print(valid_proof)
+
         valid_proof_value = HexaryTrie.get_from_proof(trie.root_hash, valid_proof_key, valid_proof)
         assert valid_proof_value == trie.get(valid_proof_key)
 
     for invalid_proof_key in deleted:
-        invalid_proof = trie.get_proof(invalid_proof_key)
-        assert len(invalid_proof) == 0
+        with pytest.raises(KeyError):
+            trie.get_proof(invalid_proof_key)
 
 
 class KeyAccessLogger(dict):
--- a/tests/test_hexary_trie.py
+++ b/tests/test_hexary_trie.py
@@ -130,6 +130,17 @@ def test_trie_using_fixtures(name, updates, expected, deleted, final_root):
     actual_root = trie.root_hash
     assert actual_root == final_root
 
+    for valid_proof_key in expected:
+        valid_proof = trie.get_proof(valid_proof_key)
+        assert len(valid_proof) > 0
+        print(valid_proof)
+        valid_proof_value = HexaryTrie.get_from_proof(trie.root_hash, valid_proof_key, valid_proof)
+        assert valid_proof_value == trie.get(valid_proof_key)
+
+    for invalid_proof_key in deleted:
+        invalid_proof = trie.get_proof(invalid_proof_key)
+        assert len(invalid_proof) == 0
+
 
 class KeyAccessLogger(dict):
     def __init__(self, *args, **kwargs):
--- a/trie/hexary.py
+++ b/trie/hexary.py
@@ -159,7 +159,10 @@ class HexaryTrie:
 
         proof = []
         verified = self._get_proof(node, trie_key, proof)
-        return tuple(proof) if verified else ()
+        if verified:
+            return tuple(proof)
+        else:
+            raise KeyError("Key does not exist")
 
     def _get_proof(self, node, trie_key, proof):
         proof.append(node)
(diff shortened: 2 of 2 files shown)


zsaladin:
There are several failures in tests.
I checked it. And I think it results from pytest version conflict.

zsaladin:
@pipermerriam Could you review this?

carver:
After #81 is merged, you can rebase to fix the tests.

trie/hexary.py
```diff
@@ -141,16 +141,59 @@ def _delete(self, node, trie_key):
...
             raise BadTrieProof("Missing proof node with hash {}".format(e.args))
 
+    def get_proof(self, key):
+        validate_is_bytes(key)
+
+        node = self.get_node(self.root_hash)
+        trie_key = bytes_to_nibbles(key)
+
+        proof = []
+        verified = self._get_proof(node, trie_key, proof)
```
carver: Raising a `KeyError` here instead would prevent silent/confusing errors when people forget to check the return value.

(I know this isn't the standard in this repository yet, but it's the direction we ought to head)

trie/hexary.py
```diff
@@ -141,16 +141,59 @@ def _delete(self, node, trie_key):
...
+        return tuple(proof) if verified else ()
+
+    def _get_proof(self, node, trie_key, proof):
+        proof.append(node)
+
+        node_type = get_node_type(node)
+        if node_type == NODE_TYPE_BLANK:
+            return False
+
+        if node_type == NODE_TYPE_LEAF:
```
carver: `elif` branches here would help to easily visualize the flow that these are alternatives

trie/hexary.py
```diff
@@ -141,16 +141,59 @@ def _delete(self, node, trie_key):
...
+
+        if node_type == NODE_TYPE_BRANCH:
+            if not trie_key:
+                return True
+
+            node = self.get_node(node[trie_key[0]])
+            trie_key = trie_key[1:]
+            return self._get_proof(node, trie_key, proof)
+
+        raise Exception("Invariant: This shouldn't ever happen")
```
carver: Then this could be in an `else` block, which helps show which invariant is being violated (`node_type` wasn't an expected value).

trie/hexary.py
```diff
@@ -141,16 +141,59 @@ def _delete(self, node, trie_key):
     #
     @classmethod
     def get_from_proof(cls, root_hash, key, proof):
-        trie = cls({})
-
+        db = {}
         for node in proof:
-            trie._persist_node(node)
+            encoded_node = encode_raw(node)
+            encoded_node_hash = keccak(encoded_node)
```
carver: This `get_from_proof` change is subtle, but potentially meaningful. It skips over a lot of logic happening in `_persist_node`. It also couples the implementation to the underlying database structure.

If you remove this change, are the tests failing? Or is this only related to a separate desire to store keys that are < 32 bytes? (almost everywhere else we are hashing before using a key, so short keys are not a problem)

If you can yank this out into another PR (with tests), I think it would speed this PR along.

tests/test_hexary_trie.py
```diff
@@ -130,6 +130,17 @@ def test_trie_using_fixtures(name, updates, expected, deleted, final_root):
     actual_root = trie.root_hash
     assert actual_root == final_root
 
+    for valid_proof_key in expected:
+        valid_proof = trie.get_proof(valid_proof_key)
+        assert len(valid_proof) > 0
+        print(valid_proof)
```
carver: Would you delete the `print()`? What's useful while building out one test becomes console spam for others.

carver:
Welcome to the repo! Thanks for the submission.

trie/hexary.py
zsaladin: I dropped it to speed this PR.

As you mentioned almost cases are covered but some cases are not.

```python
from trie import HexaryTrie

t = HexaryTrie({})
t[b'1'] = b'11'

HexaryTrie.get_from_proof(t.root_hash, b'1', [t.root_node])  # BadTrieProof
```

I am worried that failing this kind of simple codes cause confusing.

carver:
I started to review, then realized that my thoughts on it weren't much shorter than doing the code change. So I pushed my suggestions to your branch.

I'll give you some time to review and decide if you want to keep the changes or rewrite them. If you're good as is, then we can merge. Otherwise, we can continue. The gist of the two comments is:
- avoid mutable arguments (especially avoid *relying* on the arguments being mutable to pass information back from the method)
- avoid reassigning values to the same variable (like reassigning `node =` in this case)

zsaladin:
Thanks. Your suggestion is very reasonable.

- In most cases using immutable variables are better than mutable variable if we can use immutable variables. 
- `_get_proof` itself raises an exception so that `_get_proof` can return proof as immutable variable. `get_proof` does not need to check what `_get_proof` returns anymore like `is_verified`.
- Reassigning values to the same variable could cause confusing.
- KeyError exception message has became clearer. By keeping `trie_key` and adding `proven_len` make it.

I think it is okay to merge it.

carver (approved):
Great, thanks for the contribution!

PR #81, opened 2019-01-22
Remove explicit pytest requirement

carver:
### What was wrong?

`pytest-xdist` has specific `pytest` requirements

### How was it fixed?

Remove pinned `pytest` requirement, allow `pytest-xdist` to choose which `pytest` to install.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,7 +5,6 @@ from setuptools import setup, find_packages
 
 extras_require = {
     'test': [
-        "pytest==3.0.7",
         "pytest-xdist",
         "tox>=2.6.0,<3",
         "hypothesis==3.7.0",


pipermerriam (approved):
I'm a little anxious about the *implicit* `pytest` requirement.  Maybe lets actually pin down `pytest-xdist` and then pin `pytest` within a reasonably broad range?  Whatever you choose is fine.

carver:
> I'm a little anxious about the _implicit_ `pytest` requirement. Maybe lets actually pin down `pytest-xdist` and then pin `pytest` within a reasonably broad range? Whatever you choose is fine.

I get it, and I'd feel similarly if `pytest` was a core requirement. If something goes wrong (like someone removes `pytest-xdist` and forgets to re-add `pytest`, it will cleanly break in CI (but not actually affect production code). So, I'm still thinking the loose spec is fine.

PR #82, opened 2019-01-28
Could not be added a node less than 32 bytes in `get_from_proof` 

zsaladin:
### What was wrong?
Test codes. I used parts of README.md
```python
from trie import HexaryTrie

t = HexaryTrie(db={})
t.set(b'my-key', b'some-value')

HexaryTrie.get_from_proof(t.root_hash, b'my-key', [t.root_node])  # BadTrieProof
```
It was mentioned in #80 

### How was it fixed?
As I understand it, 'proof' of a node consists of path nodes to the node from root node. So in my opinion we can push the path nodes to trie made temporary for proof without additional logic.
In the current codes `_persiste_node` prevents inserting path nodes less than 32 bytes. I think there are potential risks even if it covers most cases.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/tests/test_hexary_trie.py
+++ b/tests/test_hexary_trie.py
@@ -102,6 +102,14 @@ def trim_long_bytes(param):
         return repr('0x' + param[:3].hex() + '...')
 
 
+def assert_proof(trie, key):
+    proof = trie.get_proof(key)
+    assert len(proof) > 0
+
+    proof_value = HexaryTrie.get_from_proof(trie.root_hash, key, proof)
+    assert proof_value == trie.get(key)
+
+
 @pytest.mark.parametrize(
     'name, updates, expected, deleted, final_root',
     FIXTURES_PERMUTED,
@@ -115,6 +123,7 @@ def test_trie_using_fixtures(name, updates, expected, deleted, final_root):
             del trie[key]
         else:
             trie[key] = value
+            assert_proof(trie, key)
 
     for key in deleted:
         del trie[key]
@@ -131,11 +140,7 @@ def test_trie_using_fixtures(name, updates, expected, deleted, final_root):
     assert actual_root == final_root
 
     for valid_proof_key in expected:
-        valid_proof = trie.get_proof(valid_proof_key)
-        assert len(valid_proof) > 0
-
-        valid_proof_value = HexaryTrie.get_from_proof(trie.root_hash, valid_proof_key, valid_proof)
-        assert valid_proof_value == trie.get(valid_proof_key)
+        assert_proof(trie, valid_proof_key)
 
     for invalid_proof_key in deleted:
         with pytest.raises(KeyError):
--- a/trie/hexary.py
+++ b/trie/hexary.py
@@ -215,7 +215,8 @@ class HexaryTrie:
             return BLANK_NODE_HASH
 
         if value is None:
-            # some nodes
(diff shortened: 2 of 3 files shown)


pipermerriam (approved):
:+1: 

Would you mind checking the existing tests for this method to see if they exercise the code path you mention with nodes that are less than 32 bytes and if it isn't, see about adding one so that we can 1) verify that this bug did previously exist (if it does) and 2) know that it is fixed.

pipermerriam (changes requested):
Meant to mark previous review with request changes (let me know what your findings are)

zsaladin:
I added a test for this. This test will fail in current codes in master branch.

carver:
Hm, looks like the new test isn't actually failing (besides lint). Seems worth double-checking that it is testing the expected thing first. Note that #80 is merged now.

zsaladin:
Oh sorry, my mistake. I will do double-checking next time.
I fixed lint issue and rebased master branch to this.

carver:
Hey, I had another idea for a direction to take this. I wanted to see what it would look like to avoid the repeated logic for saving a raw node.

Also, might as well build the proofs using the builtin tools in the tests. I just pushed it on top as a way to have the conversation. Thoughts?

carver:
Also, I found a way to use the proof tests with the existing fixtures so we can drop the hard-coded test. I confirmed that the latest version of the tests now catch the issue (when the root node encodes to fewer than 32 bytes).

zsaladin:
I like this change.The codes setting raw nodes are extracted to a method so it can be used where it is needed. And also I think removing hard coded tests is good choice!

Thanks!

tests/test_hexary_trie.py
```diff
@@ -102,6 +102,14 @@ def trim_long_bytes(param):
         return repr('0x' + param[:3].hex() + '...')
 
 
+def validate_proof(trie, key):
```
pipermerriam: *nitpick* 

Maybe `assert_proof`

trie/hexary.py
```diff
@@ -208,20 +208,31 @@ def root_node(self, value):
...
     # Utils
     #
+    def _set_raw_node(self, raw_node):
+        key, value = self._node_to_db_mapping(raw_node)
+        if key == BLANK_NODE:
+            # skip saving the blank node to DB
+            return BLANK_NODE_HASH
+
+        if value is None:
+            # some nodes
```
pipermerriam: > some nodes ....?

trie/hexary.py
```diff
@@ -208,20 +208,31 @@ def root_node(self, value):
     #
     # Utils
     #
+    def _set_raw_node(self, raw_node):
+        key, value = self._node_to_db_mapping(raw_node)
+        if key == BLANK_NODE:
+            # skip saving the blank node to DB
+            return BLANK_NODE_HASH
+
+        if value is None:
```
pipermerriam, line 217: Maybe the separation was/is intentional but this seem like a logical place to continue using `elif/else`.

pipermerriam (approved):

trie/hexary.py
carver: Yeah, since the first if block has an early return and the second two blocks are assigning the same two variables ahead of a return, it seemed like a nice logical separation.

carver:
Great, thanks for catching this corner case @zsaladin

PR #83, opened 2019-03-13
More info when database is missing a node, plus related pruning bugfix

carver:
### What was wrong?

When a node in the trie was missing, we:
1. got relatively little information back about what happened
2. would prune nodes *before* attempting to set/delete, so extra trie elements would disappear

### How was it fixed?

- Add a custom exception for a missing trie node
- Include information in the exception like the root hash, the key requested, and the hash of the node missing
- Add tests that uncovered pruning bug
- Fix pruning bug

TODO
- [x] Debug what appears to be situations where pruning is incomplete
- [x] Remove type check from custom exception (or maybe improve the error message and add it to all values?)

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/trie/hexary.py
+++ b/trie/hexary.py
@@ -20,6 +20,7 @@ from trie.constants import (
 )
 from trie.exceptions import (
     BadTrieProof,
+    MissingTrieNode,
 )
 from trie.utils.db import (
     ScratchDB,
@@ -65,9 +66,12 @@ class HexaryTrie:
         validate_is_bytes(key)
 
         trie_key = bytes_to_nibbles(key)
-        root_node = self.get_node(self.root_hash)
+        try:
+            root_node = self.get_node(self.root_hash)
 
-        return self._get(root_node, trie_key)
+            return self._get(root_node, trie_key)
+        except KeyError as exc:
+            self._raise_missing_node(exc, key)
 
     def _get(self, node, trie_key):
         node_type = get_node_type(node)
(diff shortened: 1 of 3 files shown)


carver:
Interesting, I wasn't sure circle was going to run on a draft

trie/hexary.py
```diff
@@ -81,62 +85,95 @@ def _get(self, node, trie_key):
...
+
         self._set_root_node(new_node)
 
     def _set(self, node, trie_key, value):
-        self._prune_node(node)
-
         node_type = get_node_type(node)
 
+        # node is mutable, so capture the key for later pruning now
+        prune_key = self._get_prune_key(node)
```
pipermerriam: This seems like potential for forgetting, what about something like:

```python
with self._prune_node(node):
    ...  # do things that might raise errors
```

The pruning would only happen *if* the context exits without exception.

pipermerriam (approved):

trie/hexary.py
carver: Argh, how did I miss that. Thanks!

trie/hexary.py
```diff
@@ -248,6 +230,30 @@ def root_node(self, value):
...
+            node_body = None
+
+        # Nothing to prune
+        if node_body is None:
+            yield
+        else:
+            # Prune only if no exception is raised
+            try:
+                yield
+            except:
```
pipermerriam, line 252: I don't think you need this try/except block at all (and we probably end up taking a performance hit for it).

```python
try:
    thing()
except:
    raise
else:
    something()

# equivalent to
thing()
something()
```

pipermerriam (commented):
probably should be cleaned up.

trie/hexary.py
carver: #86

PR #90, opened 2019-04-09
squash_changes() bugfixes, when missing trie nodes

carver:
### What was wrong?

`squash_changes()` was not tested for cases when underlying nodes were missing (like in a trinity beam sync).

There were a couple issues:
- It would return `None` instead of raise a `KeyError` if the underlying was missing
- If the root of the original trie was missing, and an exception was raised, then it would throw a second exception when trying to revert to the original root

### How was it fixed?

Added some tests to cover the various scenarios
Updated `ScratchDB` to handle the new cases (which is starting to look suspiciously like `BatchDB` -- at some point we should probably extract that out of py-evm to use here and elsewhere)

I also added a couple more tests that were already passing, just wanted to avoid regressions.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/tests/test_hexary_trie.py
+++ b/tests/test_hexary_trie.py
@@ -289,6 +289,78 @@ def test_hexary_trie_batch_save_drops_last_root_data_when_pruning():
     assert encode_hex(old_root_hash) in str(excinfo.value)
 
 
+def test_squash_changes_can_still_access_underlying_deleted_data():
+    db = {}
+    trie = HexaryTrie(db, prune=True)
+    trie.set(b'what floats on water?', b'very small rocks')
+    old_root_hash = trie.root_hash
+
+    with trie.squash_changes() as memory_trie:
+        memory_trie.set(b'what floats on water?', b'a duck')
+
+        # change to a root hash that the memory trie doesn't have anymore
+        memory_trie.root_hash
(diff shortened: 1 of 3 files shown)


tests/test_hexary_trie.py
```diff
@@ -289,6 +289,84 @@ def test_hexary_trie_batch_save_drops_last_root_data_when_pruning():
...
+        memory_trie.set(b'what floats on water?', b'ooooohh')
+        new_root_hash = memory_trie.root_hash
+
+        # change to a root hash that the memory trie doesn't have anymore
+        memory_trie.root_hash = middle_root_hash
+
+        with pytest.raises(MissingTrieNode):
+            memory_trie[b'what floats on water?']
+
+        # change the root hash back because yikes it's confused now
```
pipermerriam: You've got this message in two places and it isn't clear to me what it is trying to convey.

pipermerriam (approved):

tests/test_hexary_trie.py
carver: Looking back, I realize that it was just me being bothered with the idea of manually tweaking the root in the middle of the squash. It doesn't actually hurt anything to leave it modified. I removed the comment and the the `root_hash` reset line.

PR #91, opened 2019-04-17
Implemented key absence proof

### LINKED ISSUE #88: Proving absence of a key
### What was wrong?

One might like to prove that a key is *absent* from a trie with `get_proof()`. (Or maybe a separate API to avoid confusion)?

Right now it just crashes if you try to get a proof for a key that's not in the trie.

Add a new method `get_absence_proof()` (or something like it) that uses the trie nodes to prove that a key doesn't exist.

mttbernardini:
### What was wrong?

There was no way to obtain a proof of absence for a key in a HexaryTrie, although `get_from_proof` already deals correctly with absence proofs as shown in this test case:
https://github.com/ethereum/py-trie/blob/5e989eedd6ffd5c5d35d3ad1809817362e6ee181/tests/test_proof.py#L16-L18

Closes #88

### How was it fixed?

I modified the `get_proof` implementation making it never raise, but always returning the sufficient nodes to either prove a key has a specific value or is absent (= has an empty value).

The reason I implemented it this way, without any additional parameter, is because it's the job of `get_from_proof` to actually check the proof against the expected value (see this [comment](https://github.com/ethereum/py-trie/issues/88#issuecomment-483791493)).

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/trie/hexary.py
+++ b/trie/hexary.py
@@ -182,6 +182,9 @@ class HexaryTrie:
         node = self.get_node(self.root_hash)
         trie_key = bytes_to_nibbles(key)
 
+        if node == BLANK_NODE:
+            return (node,)
+
         return self._get_proof(node, trie_key)
 
     @staticmethod
--- a/trie/hexary.py
+++ b/trie/hexary.py
@@ -181,9 +181,6 @@ class HexaryTrie:
         node = self.get_node(self.root_hash)
         trie_key = bytes_to_nibbles(key)
 
-        if node == BLANK_NODE:
-            return (node,)
-
         return self._get_proof(node, trie_key)
 
     def _get_proof(self, node, trie_key, proven_len=0, last_proof=tuple()):
--- a/trie/hexary.py
+++ b/trie/hexary.py
@@ -194,13 +194,9 @@ class HexaryTrie:
 
         node_type = get_node_type(node)
         if node_type == NODE_TYPE_BLANK:
-            raise self._missing_key_error(trie_key)
+            return last_proof
         elif node_type == NODE_TYPE_LEAF:
-            current_key = extract_key(node)
-            if current_key == unproven_key:
-                return updated_proof
-            else:
-                raise self._missing_key_error(trie_key)
+            return updated_proof
         elif node_type == NODE_TYPE_EXTENSION:
             current_key = extract_key(node)
             if key_starts_with(unproven_key, current_key):
@@ -208,7 +204,7 @@ class HexaryTrie:
                 new_proven_len = proven_len + len(current_key)
                 return self._get_proof(next_node, trie_key, new_proven_len, updated_proof)
             else:
-                raise self._missing_key_error(trie_key)
+                return updated_proof
         elif node_type == NODE_TYPE_BRANCH:
             if not unproven_key:
                 return updated_proof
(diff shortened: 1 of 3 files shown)


carver:
@mttbernardini Looks like the `test_trie_using_fixtures` test also needs to be updated. Instead of expecting a `KeyError`, it will generate the proof successfully, and then we can `get_from_proof()` that the value is `b''`.

mttbernardini:
I fixed the test case. However it still fails as there's probably some edge case I'm not handling correctly.

I'm working on it.

trie/hexary.py
```diff
@@ -182,33 +181,28 @@ def get_proof(self, key):
         node = self.get_node(self.root_hash)
         trie_key = bytes_to_nibbles(key)
 
-        return self._get_proof(node, trie_key)
+        if node == BLANK_NODE:
+            return (node,)
```
carver: I think we can drop this special case handling by changing `_get_proof`. Commented below.

trie/hexary.py
```diff
@@ -182,33 +181,28 @@ def get_proof(self, key):
...
+        return self._get_proof(node, trie_key)
 
     def _get_proof(self, node, trie_key, proven_len=0, last_proof=tuple()):
         updated_proof = last_proof + (node, )
         unproven_key = trie_key[proven_len:]
 
         node_type = get_node_type(node)
         if node_type == NODE_TYPE_BLANK:
-            raise self._missing_key_error(trie_key)
+            return last_proof
```
carver, line 192: ~~If this is changed to `updated_proof` instead, then it will include the blank node in the proof.~~

_Edit: I think the right solution is to not include the `BLANK_NODE`, and fix whatever code is trying to load the blank node out of the database. (see below comment)_

carver (commented):
Yeah, looks like it's nearly there!

It is helpful to have the full commit history while it's a WIP, but for merging, let's squash this to just one commit. If you want to force-push the squashed commits, go ahead. Otherwise, I can squash and merge it after the change mentioned below.

carver:
Although on a second look, it's a little silly to have to include the blank node in the proof. We know what the blank node hash is, so if the root hash is the blank node hash, then we can just load it without looking it up in the proof. In fact, `get_node()` already handles that, here:
https://github.com/ethereum/py-trie/blob/2ac4b9c5f621969d78ace787728806295f703f96/trie/hexary.py#L284-L288

So I'm not sure why we'd need to include the blank node at all.

mttbernardini:
> Although on a second look, it's a little silly to have to include the blank node in the proof.

Indeed that's totally not needed and I just added it to comply with the assertion that a proof's size must be greater than zero. In fact when there's at least one node in the trie, the `BLANK_NODE` is __never__ added to the proof (and this addresses the revisions above).

In addition to this I'd also point out that the way `get_proof` is implemented, it also adds nodes which are `< 32` bytes when rlp serialized, which is redundant (except for the case when there's the root node only) because those are already embedded in their respective parent node (as per spec). Maybe we could just merge this for now and make another PR addressing both of these points (they seem related)? Let me know.

I'm gonna squash the commits if you prefer to have a merge-commit in the history, otherwise go ahead with github's squash ;)

carver:
> Maybe we could just merge this for now and make another PR addressing both of these points (they seem related)?

Eh, let's just wrap up this blank node issue now. I can see how you mean they're related, and if we didn't have to special-case `get_proof()` for this PR, I might agree. But we might as well fix it now and skip the hack altogether. Then come back for the proof size optimization of the short nodes.

carver:
Wonderful, thanks!

PR #93, opened 2019-08-16
Use reference counting to avoid over-pruning trie

### LINKED ISSUE #92: Hexary trie prunes when it shouldn't
* Version: 1.4.0
* Python: 3.6
* OS: debian linux

### What was wrong?
Trying to create a trie of a list of identical hashes, using make_trie_root_and_nodes from py-evm, and I get the exception: trie.exceptions.MissingTrieNode: Trie database is missing hash.

This only occurs if the HexaryTrie has prune = True. 

It looks like a leaf was pruned when it shouldn't have been.

The exception occurs on the 129th iteration of the loop.

Expected it to add the hashes to the trie database like normal so that I can get a root hash and nodes.

#### Code to reproduce the error

```py
    import rlp
    from trie import HexaryTrie
    ZERO_HASH32 = 32 * b'\x00'
    BLANK_ROOT_HASH = b'V\xe8\x1f\x17\x1b\xccU\xa6\xff\x83E\xe6\x92\xc0\xf8n\x5bH\xe0\x1b\x99l\xad\xc0\x01b/\xb5\xe3c\xb4!'
    
    
    def make_trie_root_and_nodes( items):
        return _make_trie_root_and_nodes(tuple(rlp.encode(item) for item in items))
    
    def _make_trie_root_and_nodes(items):
        kv_store = {}
        trie = HexaryTrie(kv_store, BLANK_ROOT_HASH, prune=True)
        memory_trie = trie
        for index, item in enumerate(items):
            index_key = rlp.encode(index, sedes=rlp.sedes.big_endian_int)
            memory_trie[index_key] = item
        return trie.root_hash, kv_store
    
    def test_trie_root():
        list_of_bytes = []
        for i in range(129):
            random_bytes = ZERO_HASH32
            list_of_bytes.append(random_bytes)
    
        print(_make_trie_root_and_nodes(list_of_bytes))
    
    test_trie_root()
```

carver:
### What was wrong?

Fixes #92 

There is an example of where the current pruner incorrectly removes a
necessary node hash, because it got duplicated. See the new test:
test_hexary_trie_avoid_over_pruning()

### How was it fixed?

Added an incremental reference counter, which only prunes nodes when the
number of usages drops to zero. It only works on fresh databases. With
existing databases, it would still delete required nodes.

Also, it skips an unnecessary database write, which would incorrectly
increment the reference counter.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/trie/hexary.py
+++ b/trie/hexary.py
@@ -271,9 +271,21 @@ class HexaryTrie:
             self._prune_key(prune_key)
 
     def _prune_key(self, key):
-        self.ref_count[key] -= 1
-        if not self.ref_count[key]:
-            del self.db[key]
+        new_count = self.ref_count[key] - 1
+
+        if new_count <= 0:
+            # Ref count doesn't track keys that are already in the starting database,
+            #   so ref count can go negative. Then, detect if key is in underlying:
+            #   - If so, delete it and set the refcount down to 0
+            #   - If not, raise an exception about trying to prune a node that doesn't exist
+            try:
+                del self.db[key]
+            except KeyError as exc:
+                raise ValidationError("Tried to prune key %r that doesn't exist" % key) from exc
+            else:
+                new_count = 0
+
+        self.ref_count[key] = new_count
 
     def regenerate_ref_count(self):
         new_ref_count = defaultdict(int)
--- a/trie/hexary.py
+++ b/trie/hexary.py
@@ -1,3 +1,4 @@
+from collections import defaultdict
 import contextlib
 import functools
 import itertools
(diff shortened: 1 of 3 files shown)


pipermerriam (commented):
Am I correct that this will not work unless the database is empty when given to the `Trie` class?  I.E. we cannot use this in trinity with our level db database?

carver:
> Am I correct that this will not work unless the database is empty when given to the `Trie` class? I.E. we cannot use this in trinity with our level db database?

Exactly, that's what I was trying to get at with the docstring:

> Important note about Pruning:
>
> Pruning against an existing database, with duplicate values, will
        delete important data. Only turn on pruning with a fresh database.
        If working with an existing database, use the :meth:`squash_changes`
        context manager instead of turning pruning on directly.

Any idea how to say that more clearly?

---

FWIW, not being able to prune an on-disk DB was already an issue. It's not made worse by the current PR (hopefully made a little better by at least mentioning it in a docstring). In practice, we have only ever used pruning for a fresh in-memory database, or in a context like `squash_changes()` which will handle it correctly.

pipermerriam:
I'm inclined to just deprecate and remove the pruning feature since:

1. databases that *need* pruning are the big ones.
2. I don't see a compelling use case for pruning in-memory databases

ergo... our pruning feature isn't useful and adding complexity to support it doesn't seem worth it.

carver:
> 2\. I don't see a compelling use case for pruning in-memory databases

Two come to mind:
1. It's used as part of `squash_changes()`. Since we don't have a batch-add, every time you add N values, you create N-1 state roots that you don't need anymore (plus all the downstream nodes that change). If we don't prune those out in memory, we would persist all of them to the database.
2. The tiny transaction and receipt tries are built in memory, using a pruning trie. For basically the same reason as 1, they would write a bunch of unnecessary trie nodes to disk if we didn't prune along the way.

It's possible that we might be able to hide away the `prune` keyword somehow and force `squash_changes` as the only approach to do that. I think we could cover both use cases that way, but I'm not sure how it would look to launch a pruning trie from inside `squash_changes()` (which uses the `prune` keyword internally right now).

pipermerriam:
> It's possible that we might be able to hide away the prune keyword somehow and force `squash_changes` as the only approach to do that.

This direction seems preferable.

carver:
> > It's possible that we might be able to hide away the prune keyword somehow and force `squash_changes` as the only approach to do that.
> 
> This direction seems preferable.

Cool, I noted it as not for external usage, and that it is likely to be deprecated. I also added an issue to change how pruning is enabled, to make it an internal API.

trie/hexary.py
```diff
@@ -243,7 +268,34 @@ def _prune_node(self, node):
 
         # Prune only if no exception is raised
         if should_prune:
-            del self.db[prune_key]
+            self._prune_key(prune_key)
+
+    def _prune_key(self, key):
+        self.ref_count[key] -= 1
```
pipermerriam: Maybe this should have an existence check since if the key were to not be in the database this would end up with the refcount being negative.

pipermerriam (approved):

PR #95, opened 2020-04-29
Add TrieFog and cache to track which parts of trie have been walked

carver:
### What was wrong?

To backfill state data in Trinity, we need some more tools to help scan the trie, keeping track of previous scan results. We need to be able to scan for nodes near an arbitrary key.

The Trinity implementation will look pretty similar to the walk section in [`test_trie_walk_root_change_with_cached_traverse_from`](https://github.com/ethereum/py-trie/pull/95/files#diff-988939f9c40d3b317a2fbbb13f12c236R398-R421)

### How was it fixed?

New class `TrieFog` to track the parts of the trie that have been explored.

New class `TrieFrontierCache` to keep a cache of recently explored nodes (in order to skip database lookups during trie traversal, by using `HexaryTrie.traverse_from()`).

Lots of tests for the trie fog, and tests about walking the trie looking for missing nodes.

Broadly, trie walk tests cover:
- iterating over all nodes using `traverse()` and trie fog
- iterating over all nodes using `traverse_from()` and trie fog
- iterating over some nodes using `traverse()` and trie fog, then modify the trie, then finish iterating
- iterating over some nodes using `traverse_from()` and trie fog, then modify the trie, then finish iterating

TODOs:
- [x] rebase on #98 
- [x] split out just `TrieFog` into PR, and rebase
- [x] use `TrieFog` in `trie/iter.py` to do basic trie iteration testing -- see #104 
- [x] remove all TODOs: either fix or open issue
- [x] type annotations
- [x] squash a lot
- [x] CHANGELOG

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/tests/test_fog.py
+++ b/tests/test_fog.py
@@ -1,0 +1,260 @@
+from eth_utils import ValidationError
+from hypothesis import (
+    given,
+    strategies as st,
+)
+import pytest
+
+from trie.exceptions import (
+    FullDirectionalVisibility,
+    PerfectVisibility,
+)
+from trie.fog import HexaryTrieFog
+
+
+def test_trie_fog_completion():
+    fog = HexaryTrieFog()
+
+    # fog should start with *nothing* verified
+    assert not fog.is_complete
+
+    # completing the empty prefix should immediately mark it as complete
+    empty_prefix = ()
+    completed_fog = fog.explore(empty_prefix, ())
+    assert completed_fog.is_complete
+
+    # original fog should be untouched
+    assert not fog.is_complete
+
+
+def test_trie_fog_expand_before_complete():
+    fog = HexaryTrieFog()
+
+    empty_prefix = ()
+    branched = fog.explore(empty_prefix, ((1,), (5,)))
+    assert not branched.is_complete
+
+    # complete only one prefix
+    single_prefix = branched.explore((1,), ())
+    assert not single_prefix.is_complete
+
+    completed = single_prefix.explore((5,), ())
+    assert completed.is_complete
+
+
+def test_trie_fog_expand_before_mark_all_complete():
+    fog = HexaryTrieFog()
+
+    empty_prefix = ()
+    branched = fog.explore(empty_prefix, ((1,), (5,)))
+    assert not branched.is_complete
+
+    # complete all sub-segments at once
+    completed = branched.mark_all_complete(((1,), (5,)))
+    assert completed.is_complete
+
+
+def test_trie_fog_composition_equality():
+    fog = HexaryTrieFog()
+
+    empty_prefix = ()
+    single_exploration = fog.explore(empty_prefix, ((9, 9, 9), ))
+
+    half_explore = fog.explore(empty_prefix, ((9, ),))
+    full_explore = half_explore.explore((9, ), ((9, 9), ))
+
+    assert single_exploration == full_explore
+
+
+def test_trie_fog_immutability():
+    fog = HexaryTrieFog()
+
+    fog1 = fog.explore((), ((1, ), (2, )))
+
+    fog2 = fog1.explore((1, ), ((3, ), ))
+
+    assert fog.nearest_unknown(()) == ()
+    assert fog1.nearest_unknown(()) == (1,)
+    assert fog2.nearest_unknown(()) == (1, 3)
+
+    assert fog != fog1
+    assert fog1 != fog2
+    assert fog != fog2
+
+
+@pytest.mark.parametrize('sub_segments', (
+    [(1, 2), (1, 2, 3, 4)],
+    [(1, 2), (1, 2)],
+))
+def test_trie_fog_explore_invalid(sub_segments):
+    """
+    Cannot explore with a sub_segment that is a child of another sub_segment,
+    or a duplicate
+    """
+    fog = HexaryTrieFog()
+    with pytest.raises(ValidationError):
+        fog.explore((), sub_segments)
+
+
+def test_trie_fog_nearest_unknown():
+    fog = HexaryTrieFog()
+
+    empty_prefix = ()
+    assert fog.nearest_unknown((1, 2, 3)) == empty_prefix
+
+    branched = fog.explore(empty_prefix, ((1, 1), (5, 5)))
+
+    # Test shallower
+    assert branched.nearest_unknown((0,)) == (1, 1)
+    assert branched.nearest_unknown((1,)) == (1, 1)
+    assert branched.nearest_unknown((2,)) == (1, 1)
+    assert branched.nearest_unknown((4,)) == (5, 5)
+    assert branched.nearest_unknown((5,)) == (5, 5)
+    assert branched.nearest_unknown((6,)) == (5, 5)
+
+    # Test same level
+    assert branched.nearest_unknown((0, 9)) == (1, 1)
+    assert branched.nearest_unknown((1, 1)) == (1, 1)
+    assert branched.nearest_unknown((2, 1)) == (1, 1)
+    assert branched.nearest_unknown((3, 2)) == (1, 1)
+    assert branched.nearest_unknown((3, 3)) == (5, 5)
+    assert branched.nearest_unknown((4, 9)) == (5, 5)
+    assert branched.nearest_unknown((5, 5)) == (5, 5)
+    assert branched.nearest_unknown((6, 1)) == (5, 5)
+
+    # Test deeper
+    assert branched.nearest_unknown((0, 9, 9)) == (1, 1)
+    assert branched.nearest_unknown((1, 1, 0)) == (1, 1)
+    assert branched.nearest_unknown((2, 1, 1)) == (1, 1)
+    assert branched.nearest_unknown((4, 9, 9)) == (5, 5)
+    assert branched.nearest_unknown((5, 5, 0)) == (5, 5)
+    assert branched.nearest_unknown((6, 1, 1)) == (5, 5)
+
+
+def test_trie_fog_nearest_unknown_fully_explored():
+    fog = HexaryTrieFog()
+    empty_prefix = ()
+    fully_explored = fog.explore(empty_prefix, ())
+
+    with pytest.raises(PerfectVisibility):
+        fully_explored.nearest_unknown(())
+
+    with pytest.raises(PerfectVisibility):
+        fully_explored.nearest_unknown((0,))
+
+
+def test_trie_fog_nearest_right():
+    fog = HexaryTrieFog()
+
+    empty_prefix = ()
+    assert fog.nearest_right((1, 2, 3)) == empty_prefix
+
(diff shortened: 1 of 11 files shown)


trie/test_walker.py
```diff
@@ -0,0 +1,440 @@
...
+
+from eth_typing import (
+    Hash32,
+)
+from eth_utils import (
+    to_bytes,
+    to_tuple,
+    ValidationError,
+)
+from sortedcontainers import SortedSet
```
pipermerriam: A brief look at the documentation gives me lots of confidence in this library.  If your experience continues to be positive after further depth with the libary you might add this to the tactical manual with a snippet of your experience.

carver:
So this PR looks huge, and it's not small, but... a whole lot of it is tests and commentary.

The actual new implementation is only 144 LOC in trie/fog.py according to `pygount`.

trie/fog.py
```diff
@@ -0,0 +1,284 @@
...
+class HexaryTrieFog:
+    """
+    Keeps track of which parts of a trie have been verified to exist.
+
+    Named after "fog of war" popular in video games like... Red Alert? IDK, I'm old.
+
+    Object is immutable. Any changes, like marking a key prefix as complete, will
+    return a new HexaryTrieFog object.
+    """
+    _unexplored_prefixes: SortedSet
```
pipermerriam: Depends on how well `sortedcontainers` does their typing, but this might be best typed as `typing.Set[...]` so that the object's items are also typed.

trie/fog.py
```diff
@@ -0,0 +1,284 @@
...
+        return f"HexaryTrieFog<{self._unexplored_prefixes!r}>"
+
+    @property
+    def is_complete(self) -> bool:
+        return len(self._unexplored_prefixes) == 0
+
+    def explore(
+            self,
+            old_prefix: NibblesInput,
+            sub_segments_input: Sequence[NibblesInput]) -> 'HexaryTrieFog':
```
pipermerriam: IIUC `sub_segment_input` is: The nibbles extensions from `old_prefix` that are still unexplored.

If so maybe this would benefit from a more descriptive name like `unexplored_sub_segments`  or `sub_segments_still_unexplored` or `foggy_sub_segments`

trie/fog.py
```diff
@@ -0,0 +1,284 @@
...
+        """
+        sub_segments = [Nibbles(segment) for segment in sub_segments_input]
+        new_fog_prefixes = self._unexplored_prefixes.copy()
+
+        try:
+            new_fog_prefixes.remove(old_prefix)
+        except KeyError:
+            raise ValidationError(f"Old parent {old_prefix} not found in {new_fog_prefixes!r}")
+
+        if len(set(sub_segments)) != len(sub_segments):
```
pipermerriam, line 86: If `sub_segments_input` was a `Set` type then we wouldn't have to perform this validation check.

trie/fog.py
```diff
@@ -0,0 +1,284 @@
...
+            raise ValidationError(
+                f"Got duplicate sub_segments in {sub_segments} to HexaryTrieFog.explore()"
+            )
+
+        # Add further validation that no segment is a prefix of another
+        all_lengths = set(len(segment) for segment in sub_segments)
+        if len(all_lengths) > 1:
+            for segment in sub_segments:
+                shorter_lengths = [length for length in all_lengths if length < len(segment)]
```
pipermerriam, line 100: It seems like we should be able to do *better* if we modeled the data as a tree.  IIUC the current approach is super-linear in runtime complexity.

```python
tree = {}
for segment in sub_segments:
    sub_tree = tree
    for nibble in segment:
        sub_tree.setdefault(nibble, {})
        if not isinstance(sub_tree[nibble], collections.Mapping):
            ...  # we've detected a that `segment` is a superset of `sub_tree[nibble]`
        sub_tree = sub_tree[nibble]
    # once the segment terminates, we store the full segment at that position 
    # so that we can later detect segments that are prefixes of other segments.
    sub_tree[segment[-1]] = segment
```

I think the untested psuedocode above has `O(n)` complexity.

trie/fog.py
```diff
@@ -0,0 +1,284 @@
...
+
+        :raises IndexError: if there are no foggy prefixes remaining
+        """
+        if isinstance(key_input, Nibbles):
+            key = key_input
+        else:
+            key = Nibbles(key_input)
+
+        # This will raise an IndexError if no prefixes exist
+        index = self._unexplored_prefixes.bisect(key)
```
pipermerriam, line 142: *maybe* catch and re-raise with our own exception message for better UX?

trie/fog.py
```diff
@@ -0,0 +1,284 @@
...
+            else:
+                return nearest_right
+
+    def nearest_right(self, key_input: NibblesInput) -> Nibbles:
+        """
+        Find the foggy prefix that is nearest on the right to the supplied key.
+
+        :raises IndexError: if there are no foggy prefixes to the right
+        """
+        if isinstance(key_input, Nibbles):
```
pipermerriam: *nitpick*

This could be dry'd a bit, since these four lines are repeated in the previous function.

trie/fog.py
```diff
@@ -0,0 +1,284 @@
...
+            nearest_left = self._unexplored_prefixes[index - 1]
+            nearest_right = self._unexplored_prefixes[index]
+
+            # is the left or right unknown prefix closer?
+            left_distance = self._prefix_distance(nearest_left, key)
+            right_distance = self._prefix_distance(key, nearest_right)
+            if left_distance < right_distance:
+                return nearest_left
+            else:
+                return nearest_right
```
pipermerriam, line 163: Looks like this will *prefer* the key to the right over the key to the left if their respective distance is the same.  Maybe worth noting this in the docstring.

trie/fog.py
```diff
@@ -0,0 +1,284 @@
...
+            prefix_list = ast.literal_eval(encoded_list.decode())
+            # confirm that all prefixes are valid nibbles, while building the sorted set
+            valid_prefixes = (Nibbles(prefix) for prefix in prefix_list)
+            # build the Trie Fog instance
+            deserialized_prefixes = SortedSet(valid_prefixes)
+            return cls._new_trie_fog(deserialized_prefixes)
+
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, HexaryTrieFog):
+            return False
```
pipermerriam, line 261: I would say this should raise a `TypeError` because we are probably doing something wrong if we are comparing to the wrong class.

trie/typing.py
```diff
@@ -57,9 +58,15 @@ def __repr__(self):
         return hex(self.value)
 
 
+# A user-input value, where each element will be validated as a Nibble instead of int
+NibblesInput = Sequence[int]
+
+
 class Nibbles(tuple):
```
pipermerriam: I think the should be `class Nibbles(Tuple[int, ...])` for more concrete typing.

pipermerriam (approved):

tests/test_fog.py
```diff
@@ -0,0 +1,244 @@
...
+    assert branched.nearest_unknown((5, 5, 0)) == (5, 5)
+    assert branched.nearest_unknown((6, 1, 1)) == (5, 5)
+
+
+def test_trie_fog_nearest_unknown_fully_explored():
+    fog = HexaryTrieFog()
+    empty_prefix = ()
+    fully_explored = fog.explore(empty_prefix, ())
+
+    with pytest.raises(IndexError):
```
pipermerriam: Looking at this test, I suspect we might be better served with a custom exception.  When I browsed the code it didn't stand out, but now I'm thinking about call sites where we'll have to catch this and `IndexError` doesn't make much sense to me as opposed to something more naturally descriptive like `FullyExplored`.

trie/fog.py
```diff
@@ -0,0 +1,284 @@
...
+        """
+        Convert a set of unexplored prefixes to a proper HexaryTrieFog object.
+        """
+        copy = cls()
+        copy._unexplored_prefixes = unexplored_prefixes
+        return copy
+
+    def serialize(self) -> bytes:
+        prefixes = list(self._unexplored_prefixes)
+        return f"HexaryTrieFog:{prefixes!r}".encode()
```
pipermerriam, line 242: Missed this on my first pass too.

Any reason you chose to do the encoding this way rather than an actual binary encoding of the prefixes?  Seems like it would take up significantly less space if they were all packed tightly.

pipermerriam (approved):
Those are some gnarly tests.  I didn't give them quite as deep a review as the other code.  :+1:

trie/fog.py
carver: We use APIs that aren't in `Set` like indexing to a value and `.bisect()`. I thought I might find a quick solution, then ended up building a `Protocol`, oops.

trie/fog.py
carver: For now I'm :-1: on the change, because I like that the argument can be passed in directly from a `HexaryTrieNode.sub_segments`. Making `sub_segments_input` a set would mean doing one of:
- casting `set(node.sub_segments)` at any call site
- changing `HexaryTrieNode.sub_segments` to a set, but it's a data structure that is really nice to keep immutable

trie/fog.py
carver: I'm not too worried about speed here because the whole thing is behind the `if len(all_lengths) > 1:` check, which won't actually trigger for any use cases we care about (branch, extension, or leaf nodes being explored one at a time).

I'll add a comment to that ^ effect, and open an issue to come back to later.

trie/fog.py
carver: :+1: I sped this up enough in the `Nibbles` class that we can always cast it. That drops it down to a one liner.

trie/fog.py
carver: I would be pretty surprised to get a `TypeError` at the console from `1 == 'a'`. I think the surprise of `==` raising an exception is probably worse than allowing a user to accidentally equality test against the wrong thing. In a quick search, I couldn't find anyone online that recommended raising a `TypeError` from `__eq__`.

trie/fog.py
carver: Nope, this was just fast to write. :+1: for compressing them

... and of course I need to do a nibbles terminator similar to the one in the trie, because there might be an odd number.

carver:
Whew, squashed 61 commits to 6

trie/fog.py
pipermerriam: I think what I may have been remembering that inspired me to do this was a SQLAlchemy API, but now that I'm remembering it was in the query API where you construct the query constraints like `Session.query(User).filter(User.name == 'bob')` where the `User.name == 'bob'` actually returns a non-boolean `Constraint` object which is later translated into the SQL query.  In this case it makes sense to raise exceptions when you do something like `User.name == 1234` since `1234` is not a string and thus the API doesn't make sense when you pass an object of the wrong type....

Long story short, :+1:

PR #98, opened 2020-05-02
Add prefix to MissingTrieNode exception

carver:
### What was wrong?

Some future requirements are to:
- get the prefix of the key used to navigate down to a trie node that then raises a `MissingTrieNode` exception
- traverse the trie directly to a node, using a series of nibbles

### How was it fixed?

Requires traversing down the nodes in a different way.

Includes another breaking change: Drop `KeyError` as a superclass of `MissingTrieNode`.

(This is an extraction from #95 )

TODO:
- [x] rebase on #97 to drop py3.5 support (PR uses f-strings)
- [x] ~maybe merge with the other trie `iter` utility~ (After adding `TrieFog`)
- [x] add release note about breaking changes

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/trie/typing.py
+++ b/trie/typing.py
@@ -1,0 +1,35 @@
+import enum
+
+from eth_utils import (
+    is_list_like,
+)
+
+
+class Nibble(enum.IntEnum):
+    Hex0 = 0
+    Hex1 = 1
+    Hex2 = 2
+    Hex3 = 3
+    Hex4 = 4
+    Hex5 = 5
+    Hex6 = 6
+    Hex7 = 7
+    Hex8 = 8
+    Hex9 = 9
+    HexA = 0xA
+    HexB = 0xB
+    HexC = 0xC
+    HexD = 0xD
+    HexE = 0xE
+    HexF = 0xF
+
+    def __repr__(self):
+        return hex(self.value)
+
+
+class Nibbles(tuple):
+    def __new__(cls, nibbles):
+        if not is_list_like(nibbles):
+            raise TypeError(f"Must pass in a tuple of nibbles, but got {nibbles!r}")
+        else:
+            return tuple.__new__(cls, (Nibble(maybe_nibble) for maybe_nibble in nibbles))
--- a/trie/validation.py
+++ b/trie/validation.py
@@ -1,3 +1,7 @@
+from eth_utils import (
+    is_integer,
+)
+
 from trie.constants import (
     BLANK_NODE,
     BLANK_HASH,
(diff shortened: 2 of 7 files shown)


carver:
This is a step toward improving the trie iterator. After #95 then I should be able to simplify (and improve performance) of `NodeIterator`.

trie/validation.py
```diff
@@ -18,6 +22,11 @@ def validate_length(value, length):
         raise ValidationError("Value is of length {0}.  Must be {1}".format(len(value), length))
 
 
+def validate_is_nibbles(value):
```
pipermerriam: This is where we'd benefit from having:

```python
class Nibble(enum.IntEnum):
    N0 = 0
    N1 = 1
    ...
    NE = 14
    NF = 15
```

Then we can either have things already typed as a nibble and thus not require any validation or just do `Nibble(value)` and it will raise a `TypeError` if the value is out of range.  For all other purposes, I believe each of the values for an `IntEnum` just evaluates as if it was an integer.

carver: Ok, tried it. Mixed feelings. I'll leave it in, because I'm pretty neutral on it. The biggest downside I see is that when you validate, you will get a confusing error if you try to use a byte-string instead of a tuple of nibbles for the prefix. So I ended up adding an extra tuple type check.

pipermerriam: I did a quick scan in the code but wasn't able to identify where/what you were referring to.  Would it be possible to address this by simply catching and re-raising a less confusing exception?

trie/typing.py
```diff
@@ -0,0 +1,35 @@
...
+    def __repr__(self):
+        return hex(self.value)
+
+
+class Nibbles(tuple):
+    def __new__(cls, nibbles):
+        if not is_list_like(nibbles):
+            raise TypeError(f"Must pass in a tuple of nibbles, but got {nibbles!r}")
+        else:
+            return tuple.__new__(cls, (Nibble(maybe_nibble) for maybe_nibble in nibbles))
```
carver, line 35: Hm, I don't think there's a good way to catch an exception from `Nibble` and clarify it, because the same exception would be raised by the following snippets:

```py
for n in b'A':
  Nibble(n)

for n in (0x41, ):
  Nibble(n)
```

So I just added a tiny wrapper for a series of nibbles, plus a test.

pipermerriam: I was just thinking:

```python
try:
    ...  # code that does `Nibble(value)`
except TypeError as err:
    raise TypeError("Some message that is more informative") from err
```

:+1: with whatever you think is best.  This might just be one of those ideas that I have that isn't actually useful in practice :shrug:

PR #101, opened 2020-05-06
Remove unnecessary node read

carver:
### What was wrong?

The `squash_changes` was *always* reading the root node, even if the trie was never modified. The correct behavior here is to not read out the root node from the database.

This might be relevant when building witnesses.

### How was it fixed?

Check if the root hash changed, before reading out the old root node from the database.

#### Cute Animal Picture

Cute animal picture


carver:
I verified locally that f8a4f17 fails before ab56d69 is applied.

pipermerriam (approved):

PR #102, opened 2020-05-08
Add HexaryTrie.traverse() and .traverse_from()

carver:
These are useful for inspecting trie structure (like when walking the
database to validate presence of all nodes). Traverse_from in particular
is helpful for caching, so you don't need to re-navigate from the root
every time.

TODO:
- [x] test `traverse_from` `MissingTraversalNode`
- [x] drop allowing requested_key = None
- [x] fix changelog for ^
- [x] new changelog entry
- [x] `missing_node_hash` of `MissingTrieNode` to `HexBytes` also
- [x] trie.root_node -> `MissingTraversalNode` error
- [x] test ^
- [x] CAP
- [x] issue to annotate root_node #103 
- [x] squash lots of commits

#### Cute Animal Picture

Cute animal picture *It's possible I'm the dog*

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -12,8 +12,9 @@ Breaking Changes
 - MissingTrieNode is no longer a KeyError, paving the way for eventually raising a KeyError instead
   of returning b'' when a key is not present in the trie
   https://github.com/ethereum/py-trie/pull/98
-- If traversing, instead of getting a key value, then MissingTrieNode.requested_key might be None
-  https://github.com/ethereum/py-trie/pull/98
+- If a trie body is missing when calling HexaryTrie.root_node, the exception will be
+  MissingTraversalNode instead of MissingTrieNode
+  https://github.com/ethereum/py-trie/pull/102
 
 Features
 ~~~~~~~~
@@ -22,6 +23,20 @@ Features
   from the database. This is important for other potential database layouts. The prefix may be None,
   if it cannot be determined. For now, it will not be determined when setting or deleting a key.
   https://github.com/ethereum/py-trie/pull/98
+- New HexaryTrie.traverse(tuple_of_nibbles) returns an annotated trie node found at the
+  given path of nibbles, starting from the root.
+  https://github.com/ethereum/py-trie/pull/102
+- New HexaryTrie.traverse_from(node_body, tuple_of_nibbles) returns an annotated trie node found
+  when navigating from the given node_body down through the given path of nibbles. Useful for
+  avoiding database reads when the parent node body is known. Otherwise, navigating down from
+  the root would be required every time.
+  https://github.com/ethereum/py-trie/pull/102
+- New MissingTraversalNode exception, analogous to MissingTrieNode, but when traversing
+  (because key is not available, and root_hash not available during traverse_from())
+  https://github.com/ethereum/py-trie/pull/102
+- New TraversedPartialPath exception, raised when you try to navigate to a node, but end up
+  part-way inside an extension node, or try to navigate into a leaf node.
+  https://github.com/ethereum/py-trie/pull/102
 
 Bugfixes
 ~~~~~~~~
--- a/README.md
+++ b/README.md
@@ -3,8 +3,6 @@
 [![PyPI](https://img.shields.io/pypi/v/trie.svg)](https://pypi.org/project/trie/)
 ![PyPI - Python Version](https://img.shields.io/pypi/pyversions/trie.svg)
 
-> Warning: This is an early release and is likely to contain bugs as well as
-> breaking API changes.
 > This library and repository was previously located at [pipermerriam/py-trie](https://github.com/pipermerriam/py-trie). It was transferred to the Ethereum foundation GitHub in 
 > November 2017 and renamed to `py-trie`.
 
@@ -25,7 +23,7 @@ pip install -e .[dev]
 You can run the tests with:
 
 ```sh
-py.test tests
+pytest tests
 ```
 
 Or you can install `tox` to run the full test suite.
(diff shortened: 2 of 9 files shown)


trie/hexary.py
```diff
@@ -144,6 +162,27 @@ def _traverse(self, root_hash, trie_key) -> Tuple['Node', Tuple[int, ...]]:
...
         return self._traverse_from(root_node, trie_key)
 
+    def traverse_from(self, node_prefix, node, trie_key_input: Tuple[int, ...]):
+        """
+        Find the node at the path of nibbles provided. Unspecified behavior if
+        path ends in the middle of an extension.
+
+        The trie does *not* re-verify the path/hashes from the node prefix to the node.
+        """
+        trie_key = Nibbles(trie_key_input)
```
carver, line 200: note to self: do the same thing for `node_prefix`

trie/hexary.py
```diff
@@ -144,6 +162,27 @@ def _traverse(self, root_hash, trie_key) -> Tuple['Node', Tuple[int, ...]]:
...
+
+        try:
+            node, remaining_key = self._traverse_from(node, trie_key)
+        except MissingTrieNode as exc:
+            raise MissingTrieNode(*exc.args[:3], prefix=node_prefix + exc.prefix)
+
+        # For now, traverse will raise if you try to traverse into
+        # part of an extension node, or leaf node.
+        if len(remaining_key) > 0:
+            raise Exception(f"Tried to traverse {remaining_key!r} past {node!r}")
```
carver: TODO: Maybe just return with a remaining key?

I think raising an exception is insufficient. Example: you have explored a trie from one root, knowing that you only need to explore the unmodified parts of the trie. Then the root changes and you want to continue exploring the unexplored parts. But due to possible changes, you might end up in the middle of an extension node on the next traversal.

carver: With `traverse_from` it shouldn't be an issue, since you're always navigating from a fixed root (by supplying the parent node).

With `traverse()` though, it's definitely a possibility, and I still need to think through how to handle all the cases properly.

Though... maybe TODO: add root hash to traverse_from() args, to fix the `MissingTrieNode` exception... or make the root hash optional?

pipermerriam: I'm very likely missing something but did you give any thought to merely returning a representation of the extension node for all paths that don't get you deep enough to hit one of the child nodes?

carver: So far, the latest code has an API (direction) that I'm happy with. Traversal is all about knowing where nodes are in the trie, and moving to them. So if you go somewhere that isn't a node (like the middle of an extension or leaf), then it should raise an exception, complete with the information you need to find out what you need to keep traversing (the parent extension node, and the path to that).

CHANGELOG
```diff
@@ -12,8 +12,6 @@ Breaking Changes
 - MissingTrieNode is no longer a KeyError, paving the way for eventually raising a KeyError instead
   of returning b'' when a key is not present in the trie
   https://github.com/ethereum/py-trie/pull/98
-- If traversing, instead of getting a key value, then MissingTrieNode.requested_key might be None
-  https://github.com/ethereum/py-trie/pull/98
```
carver, line 16: I decided this was too messy, and that a different exception with known values makes more sense (`MissingTraversalNode`)

trie/hexary.py
```diff
@@ -131,20 +153,88 @@ def _get(self, root_hash, trie_key):
         else:
             raise Exception("Invariant: This shouldn't ever happen")
 
-    def _traverse(self, root_hash, trie_key) -> Tuple['Node', Tuple[int, ...]]:
+    def traverse(self, trie_key_input: Nibbles) -> TrieNode:
+        """
+        Find the node at the path of nibbles provided. Unspecified behavior if
+        path ends in the middle of an extension.
+
+        :raises MissingTraversalNode: if a node body is missing from the database
```
pipermerriam, line 165: As well as if the path ends in an extension node?

trie/hexary.py
```diff
@@ -131,20 +153,88 @@ def _get(self, root_hash, trie_key):
...
+        node, remaining_key = self._traverse(self.root_hash, trie_key)
+
+        if len(remaining_key) > 0:
+            path_to_node = trie_key[:len(trie_key) - len(remaining_key)]
+            annotated_node = self._annotate_node(node)
+            raise TraversedPartialPath(path_to_node, annotated_node)
+        else:
+            return self._annotate_node(node)
+
+    @classmethod
```
pipermerriam: this could be a static method or even a standalone util function?  I don't see any access to `cls`

trie/typing.py
```diff
@@ -33,3 +39,32 @@ def __new__(cls, nibbles):
             raise TypeError(f"Must pass in a tuple of nibbles, but got {nibbles!r}")
         else:
             return tuple.__new__(cls, (Nibble(maybe_nibble) for maybe_nibble in nibbles))
+
+
+class TrieNode(NamedTuple):
```
pipermerriam: Should this be `HexaryTrieNode`?

trie/hexary.py
carver: That's the `TraversedPartialPath` one line below

carver:
Care to take another look @pipermerriam ? I feel pretty good about it, and have some more follow-up work to do, so if you are swamped, I'm also happy to fix-forward if you notice anything later.

These traversal APIs are used in both #95 and a privately reimplemented `NodeIterator`, with some hypothesis tests.

trie/hexary.py
```diff
@@ -131,20 +154,58 @@ def _get(self, root_hash, trie_key):
...
-            except InvalidNibbles:
-                requested_key_bytes = None
-
-            raise MissingTrieNode(root_hash, root_hash, requested_key_bytes, ())
+            raise MissingTraversalNode(root_hash, ())
 
         return self._traverse_from(root_node, trie_key)
 
-    def _traverse_from(self, node, trie_key) -> Tuple['Node', Tuple[int, ...]]:
+    def traverse_from(self, node, trie_key_input: Nibbles) -> HexaryTrieNode:
```
pipermerriam: type annotation for `node`?

trie/hexary.py
```diff
@@ -131,20 +154,58 @@ def _get(self, root_hash, trie_key):
         else:
             raise Exception("Invariant: This shouldn't ever happen")
 
-    def _traverse(self, root_hash, trie_key) -> Tuple['Node', Tuple[int, ...]]:
+    def traverse(self, trie_key_input: Nibbles) -> HexaryTrieNode:
```
pipermerriam, line 158: Not clear to me what `trie_key_input` is.  Also it is typed as `Nibbles` but then the first thing this function does is cast this value to `Nibbles(trie_key_input)`.  I'm not understanding what or why that is happening.

trie/utils/nodes.py
```diff
@@ -173,3 +176,43 @@ def encode_leaf_node(value):
     if value is None or value == b'':
         raise ValidationError("Value of leaf node can not be empty")
     return LEAF_TYPE_PREFIX + value
+
+
+def annotate_node(node_body):
```
pipermerriam: type hints since this is new code?

pipermerriam (approved):
Looks like `README` needs to be updated to document the new traversal APIs.

As for the actual traversal mechanism, I can't claim that I fully ingested it, but structurally/architecturally things look sound.

pipermerriam:
Another thing to think about as you get closer to cutting releases of the library is to export the `py.typed` so that the type hints are consumable from the outside.

trie/hexary.py
carver: `Nibbles` is just a nice way to validate that each int in the tuple is in `range(16)`, if someone passes in just a tuple of ints (since type hints are just a hint, not formally required). I'll add a comment about it being for validation.

trie/utils/nodes.py
carver: Okay, yeah, I'll look at how carefully I can define the node bodies. I think it's a recursive type definition, which I don't know/remember how to do.

carver:
> Another thing to think about as you get closer to cutting releases of the library is to export the `py.typed` so that the type hints are consumable from the outside.

Yeah, although turning on type checking internally to the library seems like a prerequisite before we do that, so we don't export types that are accidentally wrong. :stuck_out_tongue:

README.md
```diff
@@ -108,6 +106,44 @@ True
...
+>>> t = HexaryTrie(db={})
+>>> t.root_hash
+b'V\xe8\x1f\x17\x1b\xccU\xa6\xff\x83E\xe6\x92\xc0\xf8n[H\xe0\x1b\x99l\xad\xc0\x01b/\xb5\xe3c\xb4!'
+>>> t[b'my-key'] = b'some-value'
+>>> t[b'my-other-key']  = b'another-value'
+
+# Look at the root node:
+>>> root_node = t.traverse(())
+>>> root_node
+HexaryTrieNode(sub_segments=((0x6, 0xd, 0x7, 0x9, 0x2, 0xd, 0x6),), value=b'', suffix=(), raw=[b'\x16\xd7\x92\xd6', b'\xb4q\xb8h\xec\x1c\xe1\xf4\\\x88\xda\xb4\xc1\xc2n\xbaw\xd0\x9c\xf1\xacV\xb4Dk\xa7\xe6\xd7qf\xc2\x82'])
```
carver, line 122: Thanks for recommending the docs. Here it is. Hopefully it gives more of a feel for how the traversal API works in practice.

(Not specifically seeking feedback, just thought you might be interested)

PR #104, opened 2020-05-15
Convert Node iterator to use trie fog

carver:
### What was wrong?

There was a concern that the trie iterator was slow and (maybe more importantly) "convoluted" because it digs into a lot of trie internals. Now, the traversal APIs are available, which should make it a bit easier to follow. Also, with `HexaryTrieFog` and `TrieFrontierCache`, it is possible to speed it up a bit when iterating over all values sequentially.

### How was it fixed?

TODO
- [x] Rebase on #95 
- [x] squash commits

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/trie/iter.py
+++ b/trie/iter.py
@@ -4,7 +4,6 @@ from typing import (
 )
 
 from trie.exceptions import (
-    FullDirectionalVisibility,
     PerfectVisibility,
 )
 from trie.fog import (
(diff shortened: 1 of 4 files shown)


trie/iter.py
```diff
@@ -1,89 +1,127 @@
...
+        .. NOTE:: If you plan to iterate the full trie, use all() instead, for performance.
 
-    def _get_next(self, node):
-        node_type = get_node_type(node)
-        if node_type == NODE_TYPE_BLANK:
+        :return: key in bytes to the right of key_bytes, or None
+        """
+        key = bytes_to_nibbles(key_bytes)
+        traversed = ()
+        root = self._trie.traverse(traversed)
```
gsalgado: This feels a bit convoluted if all we want is to get the root. Isn't there a simper way?

gsalgado (approved):
The new code is easier to follow, but other than an inline comment I'm afraid I can't give you much useful feedback as it's been something like 2 years since I last looked at this code

pipermerriam (commented):
maybe go ahead and add type hints?

trie/iter.py
carver: Yes! Now that #106 is merged...

I will update it.

carver:
> maybe go ahead and add type hints?

done

PR #106, opened 2020-05-19
Annotate HexaryTrie.root_node, and remove setter

### LINKED ISSUE #103: Return annotated node from `HexaryTrie.root_node`
All other traversals return an annotated `HexaryTrieNode`, but `trie.root_node` just returns the raw body. The inconsistency seems problematic (and annotations would be nice on the root, too!)

TODOs:
- [ ] actually use `self.traverse(())` under the hood for `root_node`
- [ ] change locations that need the raw root node to either use `self.root_node.raw` or maybe a custom internal method for speed
- [ ] add a breaking change notice to the CHANGELOG

carver:
### What was wrong?

Fix #103 

### How was it fixed?

Annotate the return value, update tests that used the raw root node, and add a new test that `HexaryTrie().root_node == HexaryTrie().traverse(())`

Also removed the setter. Do we really need it? I guess for symmetry, you would want to set with an annotated node.

Finally: Updated the `traverse_from()` API to take the new annotated node instead of the raw node. It seemed silly to return an annotated node and expect a raw one back as an argument.

TODO
- [x] squash

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -31,7 +31,7 @@ Features
 - New HexaryTrie.traverse(tuple_of_nibbles) returns an annotated trie node found at the
   given path of nibbles, starting from the root.
   https://github.com/ethereum/py-trie/pull/102
-- New HexaryTrie.traverse_from(node_body, tuple_of_nibbles) returns an annotated trie node found
+- New HexaryTrie.traverse_from(node, tuple_of_nibbles) returns an annotated trie node found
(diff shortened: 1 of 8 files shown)


carver:
I think I'll rage-merge tomorrow. It's pretty small, just a couple cleanups, really.

pipermerriam (approved):

Issue #108, opened 2020-06-08
Handling a TraversedPartialPath exception is awkward


carver:
### What was wrong?

In an older `HexaryTrieFog` API, the `TraversedPartialPath` exception arguments made more sense. Now, they don't really help you use the fog API without a lot of argument fiddling.

### What did you expect it to do?

TPP can return a "simulated" node which helps cleanly deal with the case of using the fog to traverse to a path that doesn't exist anymore.

PR #109, opened 2020-06-08
Prune during exception

carver:
### What was wrong?

While working on #108, I wrote a hypothesis test that uncovered ~3 new bugs, including a really spooky one:

- in certain cases, `del trie[b'short']` would actually delete the key at `b'short-nope-long'`!
- `trie.set(key, b'')` would sometimes (always?) try to create a leaf node with an empty value. Instead, it should act exactly the same as `del trie[key]`
- when a `MissingTrieException` was raised during `squash_changes()` (really, any time `pruning=True`), there were times when it would prune away a node and then raise an exception. In other words, it would remove a trie node we need to walk the trie from before the change attempt.
- ~`NodeIterator.next()` didn't have any way to iterate to the `b''` key.~ Moved to #110 

### How was it fixed?

- fixed `key_starts_with(complete_key, partial_key)` to return `True` only if the entire `partial_key` is present in `complete_key`. Formerly, `key_starts_with((), (1, 0))` would `return True`, which I found surprising, and apparently so did whomever implemented `delete_kv_node()`.
- just dispatch `set(key, b'')` to `delete(key)`
- Re-worked pruning: instead of immediately deleting nodes, the trie queues up all the pruning events to be triggered only at the very end of a `set()` or `delete()`.

TODO:
- [x] CHANGELOG
- [x] squash commits

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/tests/test_hexary_trie.py
+++ b/tests/test_hexary_trie.py
@@ -11,6 +11,12 @@ from eth_utils import (
     text_if_str,
     to_bytes,
 )
+from hypothesis import (
+    example,
+    given,
+    settings,
+    strategies as st,
+)
 import pytest
 import rlp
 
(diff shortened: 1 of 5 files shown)


tests/test_hexary_trie.py
```diff
@@ -248,6 +254,222 @@ def test_hexary_trie_empty_squash_does_not_read_root():
...
+        should_continue = draw(st.integers(min_value=0, max_value=99))
+        if should_continue == 0:
+            break
+
+    return inserts, updates
+
+
+@given(
+    # I really wanted to build a search space that might automatically find the below scenario,
+    #   but it seems that the search space is too big.
```
carver: Out of date comment: the latest strategy finds the example!

tests/test_hexary_trie.py
```diff
@@ -248,6 +254,222 @@ def test_hexary_trie_empty_squash_does_not_read_root():
     assert len(flagged_usage_db.read_keys) == 0
 
 
+@st.composite
+def trie_updates_strategy(draw, max_size=256):
+    """
+    Try to generate keys that recur often without enforcing too much about what
+    the actual key value/shape is.
```
carver: Write more about what the strategy does. Maybe reuse it in the walk tests, too?

trie/hexary.py
```diff
@@ -412,40 +425,58 @@ def root_node(self) -> HexaryTrieNode:
...
+        finally:
+            # Reset for next set/delete
+            self._pending_prune_keys = None
+
     @contextlib.contextmanager
     def _prune_node(self, node):
         """
         Prune the given node if context exits cleanly.
         """
+        # TODO doesn't need to be a context manager anymore
```
carver: turn into regular method call

trie/hexary.py
```diff
@@ -412,40 +425,58 @@ def root_node(self) -> HexaryTrieNode:
...
-            #   - If so, delete it and set the refcount down to 0
-            #   - If not, raise an exception about trying to prune a node that doesn't exist
-            try:
-                del self.db[key]
-            except KeyError as exc:
-                raise ValidationError("Tried to prune key %r that doesn't exist" % key) from exc
+            self._ref_count[key] = new_count
+            '''
+            if new_count == 0:
+                del self._ref_count[key]
```
carver: After all tests pass, re-introduce this optimization

carver:
I *could* try to tease apart this PR, but the reason it's combined right now is that the same test uncovered all these issues.

tests/test_hexary_trie.py
```diff
@@ -248,6 +254,235 @@ def test_hexary_trie_empty_squash_does_not_read_root():
...
+                else:
+                    change_complete = True
+
+
+@given(
+    st.lists(st.tuples(st.binary(), st.binary())),
+    st.lists(st.binary()),
+)
+@example(
+    updates=[(b'\xa0', b'\x00\x00\x00\x00\x00\x00'), (b'\xa1', b'\x00\x00')],
```
pipermerriam: I'm not sure if there are useful comments that can be added to document these example cases.

trie/hexary.py
```diff
@@ -74,8 +74,15 @@ class _PartialTraversal(Exception):
     pass
 
 
+def prune_pending(fn):
+    def wrapped(trie_self, *args):
```
pipermerriam: maybe include type hints?  Also, this function should be decorated using `functools.wraps` so that the function name and docstrings stay intact.

tests/test_hexary_trie.py
carver: I figured out 2 of them. One of them I'm not sure. It was something that came up in the middle of implementing the pruning, not fixing an existing bug.

pipermerriam (approved):

PR #110, opened 2020-06-08
Make iter test much faster, run more examples

carver:
### What was wrong?

The deadline on `test_iter` was occasionally hitting >600ms in CI, so it seemed worth a doing a rewrite.

Also, hypothesis exposed a quirk: that you can't iterate to the `b''` key using `NodeIterator.next(key)`

### How was it fixed?

Let `hypothesis` generate the trie, including keys and values just as big as before. But also allow it to be smaller in many cases.

Additionally, can now iterate to `b''` by using a new `.next(key=None)` option, which returns the first possible key.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/tests/test_iter.py
+++ b/tests/test_iter.py
@@ -6,14 +6,13 @@ import pytest
 from hypothesis import (
     given,
     settings,
-    strategies,
 )
 
 from trie import HexaryTrie
 from trie.exceptions import MissingTraversalNode
 from trie.iter import NodeIterator
 from trie.utils.nodes import is_extension_node
-from .utils import make_random_trie
+from .utils import random_trie_strategy
 
 
 ROOT_PROJECT_DIR = os.path.dirname(os.path.dirname(__file__))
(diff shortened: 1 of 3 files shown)


tests/test_iter.py
```diff
@@ -47,29 +46,34 @@ def test_trie_next_prev_using_fixtures(fixture_name, fixture):
         assert nxt == iterator.next(point)
 
 
-@given(random=strategies.randoms())
-@settings(max_examples=10, deadline=500)
-def test_iter(random):
-    trie, contents = make_random_trie(random)
+@given(random_trie_strategy())
+@settings(max_examples=200)
```
gsalgado, line 50: Shouldn't we keep the deadline here and in test_iter_all?

gsalgado (approved):
LGTM, just curious why you dropped the deadline

tests/test_iter.py
carver: Since the test is faster, it completes within the default deadline of 200ms now. Might as well go with the tighter deadline now, I suppose.

tests/test_iter.py
```diff
@@ -6,14 +6,13 @@
...
     settings,
-    strategies,
 )
 
 from trie import HexaryTrie
 from trie.exceptions import MissingTraversalNode
 from trie.iter import NodeIterator
 from trie.utils.nodes import is_extension_node
-from .utils import make_random_trie
+from .utils import random_trie_strategy
```
pipermerriam, line 15: @carver this is the type of thing that I would typically relegate to `trie.tools.strategies` or something to prevent imports from the `./tests` directory.

PR #111, opened 2020-06-11
Fix walk with partial traverse

carver:
### What was wrong?

When traversing to a node, we occasionally get `TraversedPartialPath` exceptions (if the trie changed in the middle of walking it). In that case, we want some help from the exception to continue our walk. Previously, we would build the sub-segments manually, but that was error-prone, and turns out that it got broken in a refactor and we didn't notice!

### How was it fixed?

Fix the broken TPP handling in the walk tests. Add a new `simulated_node` which helps simplify the handling.

Also, add a new hypothesis strategy that is more likely to catch issues like this, by focusing on building some long extension nodes. (which is not super likely to happen in a smooth uniform random distribution of keys)

~PR is nearly ready, just waiting to merge #113 and #114 and rebase.~

TODO:
- [x] changelog
- [x] readme update
- [x] squashing
- [x] squashing (after review)

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -12,6 +12,9 @@ Features
 - Can now use NodeIterator to navigate to the empty key b'', using NodeIterator.next(key=None) or
   simply NodeIterator.next().
   https://github.com/ethereum/py-trie/pull/110
+- TraversedPartialPath has a new simulated_node attribute, which we can treat as a node that
+  would have been at the traversed path if the traversal had succeeded. See the readme for more.
+  https://github.com/ethereum/py-trie/pull/111
 
 Bugfixes
 ~~~~~~~~
--- a/README.md
+++ b/README.md
@@ -203,11 +203,14 @@ HexaryTrieNode(sub_segments=((0xb,), (0xf,)), value=b'', suffix=(), raw=[b'', b'
 >>> fog = fog.explore(prefix, node796f75722d6.sub_segments)
 HexaryTrieFog<SortedSet([(0x6,), (0x7, 0x9, 0x6, 0xf, 0x7, 0x5, 0x7, 0x2, 0x2, 0xd, 0x6, 0xb), (0x7, 0x9, 0x6, 0xf, 0x7, 0x5, 0x7, 0x2, 0x2, 0xd, 0x6, 0xf)])>
 
-# Index keys may not matter for your use case, so you can leave them out, but there's one more
-# feature to consider: we can look directionally to the right of an index for the nearest prefix.
+# Index keys may not matter for some use cases, so we can leave them out
+#   entirely, like nearest_unknown().
+# There's one more feature to consider: we can look directionally to the right
+#   of an index for the nearest prefix.
 >>> prefix = fog.nearest_right((0x7, 0x9, 0x6, 0xf, 0x7, 0x5, 0x7, 0x2, 0x2, 0xd, 0x6, 0xc))
 (0x7, 0x9, 0x6, 0xf, 0x7, 0x5, 0x7, 0x2, 0x2, 0xd, 0x6, 0xf)
 # That same index key would give a closer prefix to the left if direction didn't matter
+#   (See the difference in the very last nibble)
 >>> fog.nearest_unknown((0x7, 0x9, 0x6, 0xf, 0x7, 0x5, 0x7, 0x2, 0x2, 0xd, 0x6, 0xc))
 (0x7, 0x9, 0x6, 0xf, 0x7, 0x5, 0x7, 0x2, 0x2, 0xd, 0x6, 0xb)
 # So we traverse to this last embedded leaf node at `prefix`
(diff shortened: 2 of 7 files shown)


carver:
Waiting for #113 and #114 to merge and then rebase, then this should be ready for review.

tests/test_hexary_trie_walk.py
```diff
@@ -304,10 +330,7 @@ def test_trie_walk_root_change_with_traverse(
         except TraversedPartialPath as exc:
             # You might only get part-way down a path of nibbles if your fog is based on an old trie
             # Determine the new sub-segments that are accessible from this partial traversal
-            sub_segments = [
-                exc.nibbles_traversed + next_segment
-                for next_segment in exc.node.sub_segments
-            ]
+            sub_segments = exc.simulated_node.sub_segments
```
carver, line 333: The entire PR is essentially motivated by wanting to enable this change. Not just in a couple tests, but as the "recommended" approach for handling `TraversedPartialPath` while walking a trie. (ie~ the readme changes)

The old approach was a bit cumbersome, but it was also broken. We didn't even notice because the code path wasn't touched! So most of the testing changes are about increasing the chance of catching this bug and other similar bugs. (Which did expose other issues in PRs that are now merged)

carver:
If the PR feels a little overwhelming, each commit in order might be easier to review.

CHANGELOG
```diff
@@ -12,6 +12,9 @@ Features
 - Can now use NodeIterator to navigate to the empty key b'', using NodeIterator.next(key=None) or
   simply NodeIterator.next().
   https://github.com/ethereum/py-trie/pull/110
+- TraversedPartialPath has a new simulated_node attribute, which you can treat as a node that
```
cburgdorf: *super nit:* which ~~you~~ we can

README.md
```diff
@@ -247,14 +247,29 @@ TraversedPartialPath: Could not traverse through HexaryTrieNode(sub_segments=((0
...
 
-# We can now drive the exploration of the extension node further, by back-forming sub-segments:
->>> sub_segments = [
-      last_exception.nibbles_traversed + segment
-      for segment in last_exception.node.sub_segments
-]
+# We can now continue exploring the children of the extension node, by using an attribute on the exception:
+>>> sub_segments = last_exception.simulated_node.sub_segments
+((0x6, 0x5, 0x7, 0x9),)
+# Note that this sub-segment now carries you the rest of the way to the child
```
cburgdorf: I notice the file has mixed narrative perspective anyway but since most of it seems to be following the "we" form I'm going to point out this last grammar nit before I try to understand what's actually going on here haha.

>Note that this sub-segment now carries ~~you~~ us the rest of the way to the child of the node that ~~you~~ we only partially traversed into.

README.md
```diff
@@ -247,14 +247,29 @@ TraversedPartialPath: Could not traverse through HexaryTrieNode(sub_segments=((0
...
       last_exception = exc
 
-# We can now drive the exploration of the extension node further, by back-forming sub-segments:
->>> sub_segments = [
-      last_exception.nibbles_traversed + segment
-      for segment in last_exception.node.sub_segments
-]
+# We can now continue exploring the children of the extension node, by using an attribute on the exception:
+>>> sub_segments = last_exception.simulated_node.sub_segments
+((0x6, 0x5, 0x7, 0x9),)
```
cburgdorf, line 255: Makes sense :+1:

tests/test_exceptions.py
```diff
@@ -104,6 +112,89 @@ def test_valid_TraversedPartialPath_nibbles(valid_nibbles):
...
+
+@pytest.mark.parametrize(
+    'valid_nibbles',
+    (
+        (0, 0, 0),
+        (0xf, ) * 128,  # no length limit on the nibbles
+    ),
+)
+@pytest.mark.parametrize('key_encoding', (compute_extension_key, compute_leaf_key))
+def test_valid_TraversedPartialPath_untraversed_nibbles(valid_nibbles, key_encoding):
```
cburgdorf, line 128: odd casing for the method name but I know it's on purpose and if you happen to like that sort of thing I won't argue against it :)

tests/test_exceptions.py
```diff
@@ -104,6 +112,89 @@ def test_valid_TraversedPartialPath_nibbles(valid_nibbles):
...
+        assert simulated_node.sub_segments == ()
+        assert simulated_node.suffix == remaining_key
+        assert simulated_node.raw[0] == compute_leaf_key(remaining_key)
+        assert simulated_node.value == node_value
+    elif key_encoding is compute_extension_key:
+        assert simulated_node.sub_segments == (remaining_key, )
+        assert simulated_node.suffix == ()
+        assert simulated_node.raw[0] == compute_extension_key(remaining_key)
+    else:
+        raise NotImplementedError
```
cburgdorf: Seems a bit odd to raise a `NotImplementedError` from a test case. Is this supposed to be a placeholder for covering additional test cases in the future?

trie/exceptions.py
```diff
@@ -177,6 +195,78 @@ def node(self) -> HexaryTrieNode:
...
+        For the purposes of walking a trie, we might only be interested in the sub_segments,
+        suffix, etc, of the node -- but assuming we actually had a node immediately at the
+        requested prefix. This returns a node simulated as if that were true.
+
+        See the trie walk tests for an example of how this is used.
+        """
+        return self._simulated_node
+
+    def _make_simulated_node(self) -> HexaryTrieNode:
+        from trie.utils.nodes import (
```
cburgdorf, line 217: Any specific reason for the local import?

cburgdorf (approved):
I like how much effort you put into documenting this :+1: It's still hard to digest but that's just on my part because I haven't had to deal with the internals of tries yet (Though your walk-throughs have definitely helped to build up some knowledge)

I left a few comments inline. But nothing too important (still curious about the local import though)

tests/test_exceptions.py
carver: Yeah, I guess we usually say something like `raise Exception("Unreachable")`. The line of thinking was that the test needs to implement how to handle the different ways of encoding a key. But I'll change it to stick to the more team-standard way of doing it. :+1:

trie/exceptions.py
carver: Yeah, the `exceptions` module usually shouldn't import from the rest of the project, because it often will cause an import cycle. I *think* it actually does cycle here if I try to import at the top. Even if it doesn't though, I don't want to leave a landmine for the next contributor.

PR #112, opened 2020-06-11
Iterate more things

carver:
### What was wrong?

Want to be able to iterate over values, items, and nodes of the trie. `NodeIterator` previously could only iterate over keys.

### How was it fixed?

Iterate over nodes, then build all the other options on top of that.

Added a couple utilities to try to make "interesting" tries (largely borrowed from some parallel work on trie walking at #111 ).

Bonus repr implementation on HexaryTrie. At first, I thought I would build the trie and pass it in in the strategy, but that made `@example`s impossible. I added the repr which made it possible to copy in the example from the hypothesis log, but it was still ugly enough that I decided to refactor the tests. No harm in keeping the repr, though.

TODO
- [x] add pr number to changelog
- [x] squash changes

#### Cute Animal Picture

Cute animal picture


carver:
@cburgdorf I think you just missed the iterator code in the call today, so what better way to get familiar with it than to review it? :P I'm happy to hop on a call tomorrow morning to walk through it.

tests/test_iter.py
```diff
@@ -47,8 +55,7 @@ def test_trie_next_prev_using_fixtures(fixture_name, fixture):
 
 
 @given(random_trie_strategy())
-@settings(max_examples=200)
```
cburgdorf, line 50: Just curious, `max_examples` was dropped because iterating became faster and it isn't needed anymore to limit the examples?

trie/iter.py
```diff
@@ -37,7 +38,7 @@ def next(self, key_bytes: Optional[bytes] = None) -> Optional[bytes]:
         Find the next key to the right from the given key, or None if there is
         no key to the right.
 
-        .. NOTE:: If you plan to iterate the full trie, use all() instead, for performance.
+        .. NOTE:: If you plan to iterate the full trie, use keys() instead, for performance.
```
cburgdorf: *grammar nit* (avoiding *you*)

> .. NOTE:: If **we** plan to iterate the full trie, it is faster to use keys() instead

cburgdorf (approved):
Looks good to me! I definitely remember fiddling with tries once or twice where I wish I had these methods available (just to be able to observe better what is going on inside the trie)

tests/test_iter.py
carver: It was dropped because I remembered that 200 was the default :see_no_evil:

PR #113, opened 2020-06-15
Bugfix no-op squash_changes on a pruning trie

carver:
`squash_changes` was trying to prune the root node, when it was unchanged from the original trie. This only matters when the original trie has `prune=True`, which isn't a heavily-tested scenario. But we need it for some trie walk tests. The problem is that the temporary trie during squashing had an empty ref-count, so it didn't know that the old root was still needed. To solve the problem, we pass through the parent trie's ref-count.

This was caught when trying to walk over a trie that's pruning, and runs a squash-changes, but sometimes doesn't change anything at all (see #111). It's important to prune while walking because we want to run mid-walk trie updates. Those updates should delete access to old nodes, to simulate losing access to old trie nodes from peers that have moved on. This makes sure we don't request trie nodes that don't exist anymore during backfill-only type scenarios.

TODO:
- [x] CHANGELOG

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/tests/test_hexary_trie.py
+++ b/tests/test_hexary_trie.py
@@ -1013,3 +1013,39 @@ def test_traverse_from_partial_path(
     assert exc.nibbles_traversed == path_to_node
     assert exc.node.sub_segments == sub_segments
     assert exc.node.value == node_val
+
+
+def test_squash_a_pruning_trie_keeps_unchanged_short_root_node():
+    db = {}
+    trie = HexaryTrie(db, prune=True)
+    trie[b'any'] = b'short'
+    root_hash = trie.root_hash
+    with trie.squash_changes() as trie_batch:
+        trie_batch[b'any'] = b'short'
+        assert trie.root_hash == root_hash
+        assert root_hash in trie_batch.db
+        assert root_hash in db
+    assert trie.root_hash == root_hash
+    assert root_hash in trie.db
+    assert root_hash in db
+
+
+@pytest.mark.parametrize('prune', (True, False))
+def test_squash_a_trie_handles_setting_new_root(prune):
+    db = {}
+    trie = HexaryTrie(db, prune=prune)
+    with trie.squash_changes() as trie_batch:
+        trie[b'\x00'] = b'33\x00'
+        old_root_hash = trie.root_hash
+
+    # The ref-count doesn't get reset at the end of the batch, but the pending prune count does
+    # Make sure the logic here can handle that
+
+    with trie.squash_changes() as trie_batch:
+        trie_batch[b'\x00\x00\x00'] = b'\x00\x00\x00'
+
+        assert trie_batch.root_hash != old_root_hash
+        assert trie_batch.root_hash != trie.root_hash
+
+    assert trie.root_hash != old_root_hash
+    assert trie[b'\x00\x00\x00'] == b'\x00\x00\x00'
--- a/tests/test_hexary_trie_walk.py
+++ b/tests/test_hexary_trie_walk.py
@@ -17,14 +17,14 @@ from trie.iter import NodeIterator
 from trie.typing import Nibbles
 
 
-def _make_trie(keys):
+def _make_trie(keys, prune=False):
     """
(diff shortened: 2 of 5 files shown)


tests/test_hexary_trie.py
```diff
@@ -1013,3 +1013,39 @@ def test_traverse_from_partial_path(
...
+    trie = HexaryTrie(db, prune=True)
+    trie[b'any'] = b'short'
+    root_hash = trie.root_hash
+    with trie.squash_changes() as trie_batch:
+        trie_batch[b'any'] = b'short'
+        assert trie.root_hash == root_hash
+        assert root_hash in trie_batch.db
+        assert root_hash in db
+    assert trie.root_hash == root_hash
+    assert root_hash in trie_batch.db
```
carver: This is the assertion that was failing before the bugfix.

tests/test_hexary_trie_walk.py
```diff
@@ -17,14 +17,14 @@
 from trie.typing import Nibbles
 
 
-def _make_trie(keys):
+def _make_trie(keys, prune=False):
```
carver: I guess I should probably move this to `trie.tools.` ... `builder`?

trie/hexary.py
```diff
@@ -112,10 +112,16 @@ def __init__(self, db, root_hash=BLANK_NODE_HASH, prune=False):
...
+            if prune:
+                self._ref_count = defaultdict(int)
+            else:
+                self._ref_count = None
         else:
-            self._ref_count = None
+            if prune:
+                self._ref_count = ref_count
+            else:
+                raise ValueError("Cannot pass an existing reference count in to a non-pruning trie")
```
carver, line 124: Maybe a `ValidationError` instead?

pipermerriam: Depends.

Yes, if this is an exception that you would expect users to run into during normal use of the library (albeit maybe slight mis-use)

No, if this would only occur if 1) someone is mucking with internals or 2) it is only likely to be encountered while working on the core library code.

carver: > No, if this would only occur if 1) someone is mucking with internals or 2) it is only likely to be encountered while working on the core library code.

Yeah, I guess it's more of this one. So I'll leave it. :+1:

trie/hexary.py
```diff
@@ -781,15 +787,18 @@ def __contains__(self, key):
...
         scratch_db = ScratchDB(self.db)
         with scratch_db.batch_commit(do_deletes=self.is_pruning):
-            memory_trie = type(self)(scratch_db, self.root_hash, prune=True)
+            Trie = type(self)
+            memory_trie = Trie(scratch_db, self.root_hash, prune=True, ref_count=self._ref_count)
             yield memory_trie
 
         if self.root_hash != memory_trie.root_hash:
             try:
-                self._set_root_node(memory_trie.root_node.raw)
```
carver, line 789: This method had too much pruning-related baggage. For this situation, we need to manage things more directly, so just use `_set_raw_node` below.

carver:
Anything else you think this needs for a merge @pipermerriam ?

carver:
Hey @marcgarreau - mind doing a review? I know we didn't talk much about pruning and `squash_changes()` on the py-trie call. So I'm happy to hop on a call to go over that conceptually, and this PR specifically.

wolovim:
Sorry, just plain missed these notifications yesterday. This will be where I start this morning (in an ~hour). Would also welcome the walk-through call.

tests/test_hexary_trie.py
wolovim: Maybe a basic context manager detail, but curious why this would fail when the same assertion three lines above might've passed.

carver:
No worries, I was pretty focused on bugfixing a critical trinity syncing issue, today. Tomorrow after standup?

tests/test_hexary_trie.py
wolovim: as discussed in the call, I think the intention was `trie.db` here.

wolovim (approved):
Thanks for the walkthrough today! 🚢

PR #114, opened 2020-06-15
Fix overzealous TraversedPartialPath raised on leaves

carver:
### What was wrong?

`.traverse()` was raising a `TraversedPartialPath` with a leaf node, even if you try to traverse on a path unrelated to the leaf. Instead, if you traverse into an unrelated path, you should just get an empty node.

### How was it fixed?

Return the empty node if the traversed path doesn't match the beginning of the leaf suffix.

TODO:
- [x] changelog
- [x] rebase on #113 after it's merged

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/tests/test_hexary_trie.py
+++ b/tests/test_hexary_trie.py
@@ -1015,6 +1015,29 @@ def test_traverse_from_partial_path(
     assert exc.node.value == node_val
 
 
+def test_traverse_non_matching_leaf():
+    trie = HexaryTrie({})
+    EMPTY_NODE = trie.root_node
+
+    trie[b'\xFFleaf-at-root'] = b'some-value'
+    final_root = trie.root_node
+
+    # Traversing partway into the leaf raises the TraversedPartialPath exception
+    with pytest.raises(TraversedPartialPath):
+        trie.traverse((0xf,))
+    with pytest.raises(TraversedPartialPath):
+        trie.traverse_from(final_root, (0xf,))
+
+    # But traversing to any *non*-matching nibble should return a blank node, because no
+    #   children reside underneath that nibble. Returning the leaf with a mismatched nibble
+    #   would be a bug.
+    for nibble in range(0xf):
+        # Note that we do not want to look at the 0xf nibble, because that's the one that
+        #   should raise the exception above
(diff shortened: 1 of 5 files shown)


trie/tools/builder.py
```diff
@@ -11,6 +11,6 @@ def trie_from_keys(keys, prune=False):
     trie = HexaryTrie(node_db, prune=prune)
     with trie.squash_changes() as trie_batch:
         for k in keys:
-            trie_batch[k] = k
+            trie_batch[k] = k.rjust(minimum_value_length, b'3')
```
carver: I can add the same note as in `trie.tools.strategies`:

```py
# Flood 3's at the end of the value to make it longer. b'3' is encoded to 0x33,                        
#   so the bytes and HexBytes representation look the same. Just a convenience. 
```

Oh, I should also switch this to `ljust` so it's actually putting the `3`s at the end, instead of the beginning.

tests/test_hexary_trie.py
```diff
@@ -1015,6 +1015,29 @@ def test_traverse_from_partial_path(
...
+    with pytest.raises(TraversedPartialPath):
+        trie.traverse_from(final_root, (0xf,))
+
+    # But traversing to any *non*-matching nibble should return a blank node, because no
+    #   children reside underneath that nibble. Returning the leaf with a mismatched nibble
+    #   would be a bug.
+    for nibble in range(0xf):
+        # Note that we do not want to look at the 0xf nibble, because that's the one that
+        #   should raise the exception above
+        assert trie.traverse((nibble,)) == EMPTY_NODE
```
carver, line 1037: Before the attached bugfix, this `traverse()` line would also raise a `TraversedPartialPath` exception.

carver:
@njgheorghita would you mind reviewing? Only the last two commits are relevant, the others will get reviewed in #113 -- I'm happy to hop on a call if you want to talk it over.

njgheorghita (approved):
Sure! I looked over the last two commits, as far as I understood it everything made sense / looked clean. Duplicating the notes as you say in your comment, I think would be helpful.

PR #119, opened 2020-10-08
Relax rlp dependency

cburgdorf:
### What was wrong?

The current RLP dependency has a hard bound on `2.0.0a1`. Trinity, Py-EVM and friends need to migrate to `2.0.0a2` so we need to update the dependency. 

### How was it fixed?

Changed it to `"rlp>=1,<3"`. I guess the argument against an upper bound of `<3` is that we don't know yet if we end up making actual breaking changes to rlp 2 so it would be safer to pin it to `2.0.0a2`. On the other hand, having to cut new releases in `trie`,`py-evm`, `eth-rlp`, `eth-account` and `eth-tester` just because Trinity wants to go from `2.0.0a1` to `2.0.0a2` is unfortunate, too so maybe an upper bound of `<3` isn't actually that bad.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -49,7 +49,7 @@ setup(
         "eth-hash>=0.1.0,<1.0.0",
         "eth-utils>=1.6.1,<2.0.0",
         "hexbytes>=0.2.0,<0.3.0",
-        "rlp>=1,<=2.0.0-alpha.1",
+        "rlp>=1,<3",
         "sortedcontainers>=2.1.0,<3",
         "typing-extensions>=3.7.4,<4",
     ],


cburgdorf:
Oh and it seems [`eth-enr` uses an upper bound of `<3`](https://github.com/ethereum/eth-enr/blob/master/setup.py#L64) as well so I guess it's fine.

carver:
Two issues:
1. I think pip/setuptools will ignore prereleases if your bounds are only stable releases (1, and 3 here), so it won't even pick up the version you're wanting.
2. This change means that py-trie is asserting that the API of eth-rlp is guaranteed not to change anymore, even though it's in alpha.

Maybe 1 is fine, and it just means that the prerelease will only be used if downstream libraries pin it, until v2 is released as stable.

For 2, I'd be on board if you release eth-rlp as at least a beta. That means there will be no backwards-incompatible API changes in the v2 line of releases.

cburgdorf:
>I think pip/setuptools will ignore prereleases if your bounds are only stable releases (1, and 3 here), so it won't even pick up the version you're wanting.

Yes, I think that's fine. Only downstream apps such as Trinity will forcefully pin to one of the newer versions and choose the `[rust-backend]` extra.

>For 2, I'd be on board if you release eth-rlp as at least a beta. That means there will be no backwards-incompatible API changes in the v2 line of releases.

Just to double check, you mean [`rlp`](https://pypi.org/project/rlp/#history) not `eth-rlp` (which also exists), right? I would be on board with releasing the current `2.0.0a2` as `2.0.0b2`.

carver:
Oops, yes, `rlp`. I think the devnum will reset when you `make release bump=stage`, so it will be `2.0.0b1`.

cburgdorf:
Ok, it's out: https://pypi.org/project/rlp/2.0.0b1/

pipermerriam (approved):
https://github.com/ethereum/eth-tester/pull/201#pullrequestreview-507457016

carver (approved):

PR #121, opened 2021-04-28
Add node_type field to HexaryTrieNode

ftruzzi:
### What was wrong?

It looks like `get_node_type` only accepts a `RawHexaryNode`, but I think it makes sense for it to also accept a `HexaryTrieNode` instead of throwing an exception:

```python
>>> t = HexaryTrie(db={})
>>> t.set(b'test', b'test_value')
>>> get_node_type(t.root_node)
```
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/ftruzzi/projects/ethereum/py-trie/trie/utils/nodes.py", line 55, in get_node_type
    return NODE_TYPE_EXTENSION
trie.exceptions.InvalidNode: Unable to determine node type
```

### How was it fixed?

~~Made `get_node_type` access the raw node if a `HexaryTrieNode` is passed as an argument.~~

Added node_type field to `HexaryTrieNode`.

#### Cute Animal Picture

image

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,7 +1,11 @@
 Unreleased
 ---------------
 
-None
+Features
+~~~~~~~~
+
+- Added node_type field to HexaryTrieNode so that users can easily inspect the type
+  of a node.
 
 v2.0.0-alpha.4
 ---------------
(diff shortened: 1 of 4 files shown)


ftruzzi:
@carver mind taking another look? thanks!

carver:
Yeah, this looks great, thanks!

PR #122, opened 2021-11-19
chore: upgrade typing-extensions

antazoey:
### What was wrong?

We are experiencing weird build failures down the pipeline from this, related to typing extensions 4.0.0 that was just released a couple days ago.

### How was it fixed?

upgrading typing-extensions

#### Cute Animal Picture

toy_elephant

PR #124, opened 2021-11-30
deps: upgrade typing-extensions to v4

fubuloubu:
### What was wrong?
`typing-extensions` v4 was released on November 15th, however the version pinning precludes it's use and causes several package conflicts due to too-tight pinning

### How was it fixed?
Upgrading the package

#### Cute Animal Picture
angry panda


kclowes:
I'm going to close since I just merged #122.

PR #125, opened 2021-12-01
Support py39 py310, drop python 3.6

kclowes:
### What was wrong?

We weren't officially supporting python 3.9 or 3.10.

This update requires an `eth-utils` and `pyrlp` version requirement bump. The `eth-utils` version requirement has already been updated in `setup.py`. For `pyrlp` the needed changes are in master, so once that gets released, I'll update the version requirement here and tests will pass. 

### How was it fixed?

Added CI test runs specifically for python 3.9 and 3.10. Also cleans up some deprecation warnings. 

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -65,7 +65,6 @@ setup(
         'Natural Language :: English',
         "Operating System :: OS Independent",
         'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: 3.8',
         'Programming Language :: Python :: 3.9',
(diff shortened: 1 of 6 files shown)


carver:
After this is merged, let's push py-trie as a stable v2 release.

setup.py
```diff
@@ -47,9 +47,9 @@
 
     install_requires=[
         "eth-hash>=0.1.0,<1.0.0",
-        "eth-utils>=1.6.1,<2.0.0",
+        "eth-utils>=2.0.0,<3.0.0",
         "hexbytes>=0.2.0,<0.3.0",
-        "rlp>=1,<3",
+        "rlp>=3,<4",
```
carver, line 52: Not a backwards-compatible change, but that's okay, since v2 is in alpha.

kclowes:
Ha, I was just going to ask you what was needed for a stable release. Sounds good!

carver (approved):
Since py36 EOL is in just a couple of weeks, might as well remove that too (probably small enough to put in this PR). Then we won't need a new major version of py-trie when dropping py36 soon.

PR #127, opened 2022-03-02
Make typing-extensions optional

Uxio0:
- It's not required for Python >= 3.8

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,43 +44,64 @@ common: &common
         key: cache-v1-{{ arch }}-{{ .Environment.CIRCLE_JOB }}-{{ checksum "setup.py" }}-{{ checksum "tox.ini" }}
 
 jobs:
-  lint:
-    <<: *common
-    docker:
-      - image: circleci/python:3.6
-        environment:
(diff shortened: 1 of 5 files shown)


mmv08 (approved):

sunwrobert:
hi @kclowes , can we please merge this in? Currently running into issues when using `eth-brownie` and installing `gnosis-py` due to the `typing-extensions` dependency.

kclowes:
Thanks for the PR @Uxio0! I changed it to follow the compat pattern from web3, just to be consistent across our libraries. I also added additional lint runs for each Python version we support which feels a little overkill, but I've also run into issues with linting on different Python versions recently. I don't think it adds too much overhead but we can re-evaluate if it feels too clunky or if we want our tests to run faster later.

fselmo (approved):
@kclowes this lgtm!

kclowes:
@Uxio0 @sunwrobert I'll get this released early next week!

kclowes:
Sorry about the delay, @Uxio0 and @sunwrobert. 2.0.1 is out with this change!

PR #128, opened 2022-03-10
fix: reversed root->leaf to leaf->root

### LINKED ISSUE #129: `smt.exists(item)` doesn't work if default node value is set
https://github.com/ethereum/py-trie/blob/e5bf55c66c19bb21b626b8d61925ebc4336cba0c/trie/smt.py#L340-L341

should be this instead:

```py
 return self.get(key) == self._default
```

sabotagebeats:
### What was wrong?
reversed root->leaf to leaf->root in trie/smt.py

### How was it fixed?
removed reversed()

#### Cute Animal Picture
https://i.natgeofe.com/k/c022030e-f1aa-4ab3-ad56-fdcdd4a1d08b/125-animals-tiger_4x3.jpg

fixes #129

### DIFF
--- a/trie/smt.py
+++ b/trie/smt.py
@@ -54,9 +54,9 @@ def calc_root(key: bytes, value: bytes, branch: Sequence[Hash32]) -> Hash32:
     path = to_int(key)
     target_bit = 1
     # traverse the path in leaf->root order
-    # branch is in root->leaf order (key is in MSB to LSB order)
+    # branch is in leaf->root order (key is in MSB to LSB order)
     node_hash = keccak(value)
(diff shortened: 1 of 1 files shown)


sabotagebeats:
Please let me know if there's anything other I need to do. I may also need to fix tests.

sabotagebeats:
fixes 129

kclowes:
Thanks for the PR @sabotagebeats! Sorry it got lost in the shuffle. Without digging into this too closely, it looks like removing `reverse` is a breaking change. But it looks like the tests pass if that part is removed. To be merged, tests need to be passing, and ideally a test case would be added for the fix.

PR #134, opened 2022-11-30
Upgrade template

kclowes:
### What was wrong?
Pulled in the ethereum-python-project-template. I think this is the right way to do it, following these instructions [here](https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/template.md). Unfortunately, we need to merge the histories of both this library and the ethereum-python-project-template to be able to pull in changes easily for next time. But I'm willing to explore other options if anyone has a better idea to keep the py-trie history clean.

### How was it fixed?

Pulled it in, added black checks and isort checks

#### Cute Animal Picture

Cute animal picture


pacrob (approved):
Manual says we want the full template history, so it would look similar even if the project was started "correctly" with the template. I say :ship:

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,11 +1,19 @@
...
 
-### How was it fixed?
+Summary of approach.
 
+### To-Do
 
+[//]: # (Stay ahead of things, add list items here!)
+- [ ] Clean up commit history
+
+[//]: # (See: https://<RTD_NAME>.readthedocs.io/en/latest/contributing.html#pull-requests)
```
fselmo, line 14: This should replace `<RTD_NAME>` with `py-trie` but I don't think we have a "contributing" section or a readthedocs for this project... so we can just remove this line?

fselmo (commented):
nit change

PR #135, opened 2022-12-07
Remove reference to RTD

kclowes:
## What was wrong?
There was a stray Read the Docs reference in the Pull Request Template. 

## How was it fixed?

Removed it. 

### To-Do

[//]: # (Stay ahead of things, add list items here!)
- [x] Clean up commit history

#### Cute Animal Picture

put a cute animal picture link inside the parentheses

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -6,6 +6,7 @@ envlist=
 [isort]
 combine_as_imports=True
 force_sort_within_sections=True
+force_grid_wrap=1
 known_third_party=hypothesis,pytest
 profile=black
 multi_line_output=3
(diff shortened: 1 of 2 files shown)


kclowes:
@fselmo I also made some tweaks to the isort config in the second commit, mostly because I wanted to make sure that these pass irl, before I stick them in the template.

tox.ini
```diff
@@ -6,9 +6,10 @@ envlist=
 [isort]
 combine_as_imports=True
 force_sort_within_sections=True
+force_grid_wrap=1
 known_third_party=hypothesis,pytest
-profile=black
```
fselmo, line 10: nit: Is moving this necessary? might just messy up some commit history. I'm fine with it though.

fselmo (approved):
lgtm, I like the config approach. Just added the tiniest of nits which is fine if left alone too 😄

tox.ini
kclowes: Ha, nope. Not sure why that got moved 🤔

PR #137, opened 2023-05-12
Upgrade template

pacrob:
## What was wrong?

The project template was recently updated. This pulls in those changes, except for `mypy` and `pydocstyle`.

### To-Do

- [x] Add newsfragment

#### Cute Animal Picture

image


kclowes (commented):
I noticed that the core tests on CI are only running the one test in `tests/core` - will you just make sure that has always been the case and that the others are being run somewhere else  before merging?

pacrob:
> I noticed that the core tests on CI are only running the one test in `tests/core` 

That's a pattern I need to watch for - the template expects tests to be in subfolders, but some libraries just have the test files directly in `tests`. I've updated `tox.ini` to run them all.

kclowes (approved):
Nice! 🎉

PR #131, opened 2022-09-15
Update ``ethereum/tests`` fixture to v11.1 and update .gitignore

fselmo:
Update `ethereum/tests` to latest `v11.1` release commit hash

#### Cute Animal Picture

Cute animal picture

PR #143, opened 2023-10-27
Fix failing tests after updating fixtures

### LINKED ISSUE #126: Update ethereum/tests fixture
* Version: v2.0.0

### What was wrong?
The `ethereum/tests` fixture hasn't been updated in a long time. It doesn't look like there will need to be major changes, just some expectation changes (famous last words).

fselmo:
### What was wrong?

closes #126
closes #131 

- Update fixture to ``v12.4`` (commit hash [6613563](https://github.com/ethereum/tests/releases/tag/v12.4))
- It looks as though the fixtures submodule hadn't been updated in a long time. There were failing tests but they did not reflect issues with the library, only issues in the test setup. The order of the updates was not being preserved for some tests where this is crucial. These changes only perform permutations on tests where there aren't updates to the same key value and the key still exists in the final state (is not deleted). For those tests, we have to preserve the update order.

### Todo:
- [x] Clean up commit history
- [x] Add entry to the [release notes](https://github.com/ethereum/py-trie/blob/master/newsfragments/README.md)

#### Cute Animal Picture

Put a link to a cute animal picture inside the parenthesis-->

### DIFF
--- a/newsfragments/143.internal.rst
+++ b/newsfragments/143.internal.rst
@@ -1,0 +1,1 @@
+Update `ethereum/tests` fixture to ``v12.4``.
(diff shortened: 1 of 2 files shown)


reedsa (approved):
Nice! Lgtm

PR #148, opened 2023-12-04
minor updates from the project template

pacrob:
### What was wrong?

minor updates from the project template, including ignore .bumpversion.cfg in pre-commit

I don't think this needs a newsfragment, but can add if requested.

### Todo:

- [x] Clean up commit history

- [x] Add or update documentation related to these changes

- [ ] Add entry to the [release notes](https://github.com/ethereum/py-trie/blob/main/newsfragments/README.md)

#### Cute Animal Picture

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,7 @@ notes: check-bump
 
 release: check-bump clean
 	# require that upstream is configured for ethereum/py-trie
-	git remote -v | grep "upstream\tgit@github.com:ethereum/py-trie.git (push)\|upstream\thttps://github.com/ethereum/py-trie (push)"
+	@git remote -v | grep -E "upstream\tgit@github.com:ethereum/py-trie.git \(push\)|upstream\thttps://(www.)?github.com/ethereum/py-trie\(push\)"
 	# verify that docs build correctly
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -53,7 +53,7 @@ notes: check-bump
 
 release: check-bump clean
 	# require that upstream is configured for ethereum/py-trie
-	git remote -v | grep "upstream\tgit@github.com:ethereum/py-trie.git (push)\|upstream\thttps://github.com/ethereum/py-trie (push)"
+	@git remote -v | grep -E "upstream\tgit@github.com:ethereum/py-trie.git \(push\)|upstream\thttps://(www.)?github.com/ethereum/py-trie\(push\)"
```
kclowes, line 56: Just out of curiosity, what does the `-E` flag do?

kclowes (approved):
LGTM!

Makefile
pacrob: extended regex - it was to make the same grep pattern work on mac and linux, I believe.

fselmo: It added the ability to use `(www.)?` depending on whether you `git clone`'d the repo with or without the `www.` in the path

PR #151, opened 2024-03-22
bump to hexbytes>=1.2.0

pacrob:
### What was wrong?

The previous capping of `hexbytes` dep was unneeded as it wasn't actually breaking for this lib.

### How was it fixed?
Opened to `hexbytes>=0.2.3` - the first version that tested against `>=py38`

### Todo:

- [x] Clean up commit history

- [x] Add or update documentation related to these changes

- [x] Add entry to the [release notes](https://github.com/ethereum/py-trie/blob/main/newsfragments/README.md)

#### Cute Animal Picture

image

### DIFF
--- a/newsfragments/151.internal.rst
+++ b/newsfragments/151.internal.rst
@@ -1,0 +1,1 @@
+Open ``hexbytes`` dep to ``>=0.2.3``
--- a/setup.py
+++ b/setup.py
@@ -48,7 +48,7 @@ setup(
     install_requires=[
         "eth-hash>=0.1.0",
         "eth-utils>=2.0.0",
-        "hexbytes>=0.2.0,<0.4.0",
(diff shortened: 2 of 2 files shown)


kclowes (commented):
Can we just leave this as `hexbytes>=0.2.0`?

pacrob:
I suppose so, though `v0.2.0` supported py37 as a max, which none of our libs officially support.

If the goal is to keep pins as low as possible without breaking, that then calls https://github.com/ethereum/web3.py/issues/3239 into question too.

kclowes:
I can see either way - leaving it as `0.2.0` or bumping to `0.2.3`. It doesn't look like we're enforcing <3.7 in `0.2.0`, we're just not testing against it.

kclowes (approved):
🎉

